PR #353, opened 2021-06-08
fix: retrieve package names to their respective packages

WillieRuemmele:
### What does this PR do?
when retrieving with packagenames, it will place their metadata into their respective package.  The location of the package metadata can also be changed by defining `PackageOptions.outputDir` as part of the retrieve options, which is what the VS Code extension would use.

NOTE: this is a breaking change and will require a major version bump since the name of `RetreiveOptions.packageNames` was changed to `RetrieveOptions.packages`, which is a better name given the changed type of that option.

@W-9252107@

### Functionality Before
before all the metadata was placed under the default package

<insert gif and/or summary>

### Functionality After
the metadata is placed into the expected package dir
<insert gif and/or summary>

### DIFF
--- a/src/client/metadataApiRetrieve.ts
+++ b/src/client/metadataApiRetrieve.ts
@@ -16,6 +16,8 @@ import {
   RetrieveOptions,
   MetadataTransferResult,
   RetrieveRequest,
+  PackageOptions,
+  RetrieveExtractOptions,
 } from './types';
 import { MetadataTransfer, MetadataTransferOptions } from './metadataTransfer';
 import { MetadataApiRetrieveError, MissingJobIdError } from '../errors';
@@ -25,11 +27,6 @@ import { asBoolean, isString } from '@salesforce/ts-types';
 
 export type MetadataApiRetrieveOptions = MetadataTransferOptions &
   RetrieveOptions & { registry?: RegistryAccess };
-
-interface Package {
-  name: string;
-  zipTreeLocation: string;
-}
 
 export class RetrieveResult implements MetadataTransferResult {
   public readonly response: MetadataApiRetrieveStatus;
@@ -146,9 +143,9 @@ export class MetadataApiRetrieve extends MetadataTransfer<
   }
 
   protected async pre(): Promise<AsyncResult> {
-    const { packageNames } = this.options;
+    const { packages } = this.options;
 
-    if (this.components.size === 0 && (!packageNames || packageNames.length === 0)) {
+    if (this.components.size === 0 && !packages?.length) {
       throw new MetadataApiRetrieveError('error_no_components_to_retrieve');
     }
 
@@ -160,8 +157,8 @@ export class MetadataApiRetrieve extends MetadataTransfer<
 
     // if we're retrieving with packageNames add it
     // otherwise don't - it causes errors if undefined or an empty array
-    if (packageNames) {
-      requestBody.packageNames = packageNames;
+    if (packages?.length) {
+      requestBody.packageNames = this.getPackageNames();
     }
 
     // @ts-ignore required callback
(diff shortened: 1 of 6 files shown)


src/client/metadataApiRetrieve.ts
```diff
@@ -177,28 +182,41 @@ export class MetadataApiRetrieve extends MetadataTransfer<
...
+    const tree = await ZipTreeContainer.create(zip);
+    // initialize with the specified output and unpackaged metadata
+    const packageAggregator: Package[] = [{ name: output, zipTreeLocation: 'unpackaged' }];
+    packageNames?.forEach((pkg) => {
+      // add the retrieved packages - if they exist
+      packageAggregator.push({ name: pkg, zipTreeLocation: pkg });
+    });
+    const validatedPackages = packageAggregator.filter((pkg) =>
+      tree.isDirectory(pkg.zipTreeLocation)
+    );
```
shetzel, line 197: I think we should keep lines 185-188, but I'm not liking the packageAggregator and validated packages code.  We should revert most of that back to how it was originally.  So I think the general flow for this function should be to handle unpackaged metadata, then if there are packageNames defined in the options, handle the packages and aggregate the components.  The `outputConfig` should only be built once since it won't change.

From the previous code, I think we can just change line 193 to: `fsPaths: ['unpackaged']`

Then iterate over any packageNames and build componentSets per package and aggregate it all and convert.

test/client/metadataApiRetrieve.test.ts
```diff
@@ -68,27 +69,24 @@ describe('MetadataApiRetrieve', async () => {
         }
       });
 
-      it('should call retrieve with given options', async () => {
+      it('should call retrieve with for unpackaged data', async () => {
```
shetzel, line 72: I think we should have tests for:

1. retrieving unpackaged
2. retrieving a package
3. retrieving multiple packages
4. retrieving a package with unpackaged

And we should verify the component paths are correct.  Also please make sure the names of the tests make sense without silly grammar errors like "with with".  :)

shetzel (changes requested):

src/client/metadataApiRetrieve.ts
shetzel: Meh...not really liking my comments here.  I was wrong about `outputConfig` - that changes per package retrieved so ignore that.  I think the iteration is fine and afaict it needs to happen that way since the outputConfig, zipComponents, and conversion is computed per package retrieved.  I'm not sure what `validatedPackages` means but seems like that could be moved into the `forEach` above.  There will always be an `unpackaged` dir in the zip due to the manifest.  Maybe change `packageAggregator` array name to `packages`.

brpowell:
I might be doing something wrong here, but I tried this and the packaged data wasn't being saved in a new directory named after itself:

```typescript
// retrieving some unpackaged data as well
const oneClass = '/Users/b.powell/dev/dx-projects/dreamhouse-lwc/force-app/main/default/classes/MyClass.cls'
const set = ComponentSet.fromSource(oneClass);

const retrieve = await set.retrieve({
  usernameOrConnection: 'user@example.com',
  output: '/Users/b.powell/dev/dx-projects/dreamhouse-lwc/force-app',
  merge: true,
  // retrieve everything in MyPkg package
  packageNames: ['MyPkg']
});

await retrieve.pollStatus();
```

The unpackaged data is being retrieved correctly but none of the other metadata:

Screen Shot 2021-06-28 at 4 02 14 PM

WillieRuemmele:
@brpowell the word package here is referring to package installed in your org, aka a `force:package:install` package. You can install this package `{ id: '04t6A000002zgKSQAY', name: 'ElectronBranding' }` and retrieve it via `--packagenames ElectronBranding` and it'll create a new folder

brpowell:
Screen Shot 2021-06-28 at 4 24 59 PM

@WillieRuemmele I have two packages in my org and when I specify to retrieve them, I can see the metadata in the zip file when stepping through the debugger. I think what I'm hitting is the metadata is being retrieved to the current working directory of the process and at the moment there isn't a way to control this. It's okay for the CLI use case where users are forced to run it in an SFDX project, but this wouldn't work for VS Code for instance. The CWD of the node process isn't the project in that context.

WillieRuemmele (commented):
Are we doing another major version bump for this breaking change?

src/client/types.ts
```diff
@@ -232,20 +232,48 @@ export interface MetadataApiRetrieveStatus {
...
 
+export interface PackageOptions {
+  /**
+   * The name of the package to retrieve.
+   */
+  name: string;
+  /**
+   * The directory where the retrieved package source should be
+   * converted. If this is not specified the directory will
+   * default to `<process.cwd()>/packageName`.
```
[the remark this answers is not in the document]
shetzel: Yes good point.  I'll change it to `PackageOptions.name` to be super clear.  👍

shetzel (approved):

test/mock/client/transferOperations.ts
```diff
@@ -195,7 +196,19 @@ export async function stubMetadataRetrieve(
             type: success.type.name,
             fileName: content,
           });
-          zipEntries.push(join('unpackaged', content));
+          if (!packages) {
+            zipEntries.push(join('unpackaged', content));
+          } else {
+            let pkgs: string[] = [];
+            if (typeof packages[0] === 'string') {
+              pkgs = packages as string[];
```
xyc, line 204: maybe `packageNames` is a better name

src/client/types.ts
```diff
@@ -232,20 +232,48 @@ export interface MetadataApiRetrieveStatus {
...
-   * Whether or not to merge and replace input components with the retrieved versions
+   * Whether or not to merge and replace input components with the retrieved versions.
    */
   merge?: boolean;
   /**
-   * A list of package names to retrieve
+   * A list of package names to retrieve, or package names and their retrieval locations.
    */
-  packageNames?: string[];
+  packages?: string[] | PackageOptions[];
```
xyc, line 276: For the sake of grokkability maybe `packageOptions` is a better name for this. Also, semantically it looks cleaner to define this as a union type `type PackageOptions = string | { ... }`.

test/mock/client/transferOperations.ts
shetzel: Sure I can change that.

src/client/types.ts
shetzel: A PackageOption is `{ name: string; outputDir?: string }` while `string[]` is an array of package names, which doesn't seem like an option to me.  The internal code needs to distinguish between the 2 so they really aren't 1 thing.  I.e., "if packages is string[] then cast to string[] and do this, but if not then cast to PackageOption[] and do this."  If it's all a `PackageOption[]` then we'd need to come up with another subtype to refer to `{ name: string; outputDir: string }[]`.

The js doc makes it pretty clear what the types mean but we could do this and it would cause less refactoring:
type PackageNames = string[]
type PackageOptions = { name: string; outputDir: string }[]
...
packages?: PackageNames | PackageOptions;

xyc (approved):

PR #361, opened 2021-06-23
fix: bump the version of @salesforce/core for PollingClient fix

shetzel:
### What does this PR do?
Bump the version of @salesforce/core to 2.24.1 to get a fix for the PollingClient.  Fix and write more unit
tests to guard against regressions.

@W-9505934@

### Functionality Before
The PollingClient from the `@salesforce/core` library had a bug that was not awaiting the result of the polling function before the setInterval timer would call the polling function again.  This was causing larger metadata retrieves to throw an error because the request would complete but the polling would send another request to the server and that extra request(s) would cause an error.

### Functionality After
The fixed PollingClient in the core library now properly waits for the polling function to return before polling again.

### DIFF
--- a/test/client/metadataTransfer.test.ts
+++ b/test/client/metadataTransfer.test.ts
@@ -4,7 +4,7 @@
  * Licensed under the BSD 3-Clause license.
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
-import { createSandbox, SinonFakeTimers, SinonStub } from 'sinon';
+import { createSandbox, SinonStub } from 'sinon';
 import { MockTestOrgData, testSetup } from '@salesforce/core/lib/testSetup';
 import { ComponentSet } from '../../src';
 import { MetadataTransfer } from '../../src/client/metadataTransfer';
@@ -18,6 +18,7 @@ import { expect } from 'chai';
 import { MetadataTransferError } from '../../src/errors';
 import { mockConnection } from '../mock/client';
 import { fail } from 'assert';
+import { sleep } from '@salesforce/kit';
 
 const $$ = testSetup();
 const env = createSandbox();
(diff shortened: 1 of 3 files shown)


test/client/metadataTransfer.test.ts
```diff
@@ -170,6 +169,27 @@ describe('MetadataTransfer', () => {
       expect(listenerStub.callCount).to.equal(1);
     });
 
+    it('should wait for polling function to return before queing another', async () => {
```
shetzel, line 172: This sets the polling frequency to 20 ms but the checkStatus function returns in 50 ms.  With the old PollingClient the callOrder array would be ['firstCall1', 'secondCall1', 'firstCall2'] but with the fix it's ['firstCall1', 'firstCall2', 'secondCall1'].

test/client/metadataTransfer.test.ts
```diff
@@ -211,51 +235,28 @@ describe('MetadataTransfer', () => {
 
       expect(checkStatus.notCalled).to.be.true;
       expect(result).to.deep.equal({ id: '1' });
+      expect(cancelListenerStub.callCount).to.equal(1);
+      expect(updateListenerStub.callCount).to.equal(0);
```
shetzel, line 239: Ensures checkStatus is not called with an immediate cancel.  Also ensures that the cancel listener is called while the update listener is not.

test/client/metadataTransfer.test.ts
```diff
@@ -211,51 +235,28 @@ describe('MetadataTransfer', () => {
...
       });
 
-      await operationPromise;
+      await operation.start();
+      const result = await operation.pollStatus();
 
+      expect(checkStatus.calledOnce).to.be.true;
+      expect(result).to.deep.equal({ id: '1' });
       expect(cancelListenerStub.callCount).to.equal(1);
+      expect(updateListenerStub.callCount).to.equal(1);
```
shetzel, line 259: Ensures checkStatus is only called once since it's canceled within the checkStatus function.  Update and cancel listeners are called once.
The old test isn't really necessary anymore since polling is exited based on the `done` property rather than `status`.  There is coverage of polling stoppage in the metadataDeployApi tests, as the coverage numbers indicate.

brpowell (approved):
Awesome getting the polling fix in. If you wouldn't mind giving PDT a heads up that core will need to be bumped that would be great :) thanks!

PR #366, opened 2021-06-30
fix: add WaveComponent to metadata registry

smithgp:
### What does this PR do?

Add `WaveComponent` to metadata registry here so deploy and retrieve work.
`WaveComponent` works like `WaveDashboard` since it's an extension of `WaveDashboard`.

@W-9520846@

### Functionality Before

Currently, deploy and retrieve on `WaveComponent` returns error from this library and plugins-source.

### Functionality After

It works, tested via `resolve('/path/to/wave/file.comp')` and when linked into plugins-source linked into sfdx.

### DIFF
--- a/src/registry/registry.json
+++ b/src/registry/registry.json
@@ -1047,6 +1047,16 @@
         "adapter": "matchingContentFile"
       }
     },
+    "wavecomponent": {
+      "id": "wavecomponent",
+      "name": "WaveComponent",
+      "suffix": "wcomp",
+      "directoryName": "wave",
+      "inFolder": false,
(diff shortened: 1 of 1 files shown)


smithgp:
@forcedotcom/pdt - This PR originally did appear to run the checks, but, when I rebased to pickup the latest, it didn't. Is there something I need to do? I can't see anything for it in circleci, maybe I don't have permissions?

shetzel (approved):
I tested these changes using the source plugin with a wave component and app.  Works great!
tested deploy and retrieve with -p and -m params.

PR #370, opened 2021-07-01
fix: add WaveComponent to metadata registry

shetzel:
What does this PR do?
Add WaveComponent to metadata registry here so deploy and retrieve work.
WaveComponent works like WaveDashboard since it's an extension of WaveDashboard.

@W-9520846@

Functionality Before
Currently, deploy and retrieve on WaveComponent returns error from this library and plugins-source.

Functionality After
It works, tested via resolve('/path/to/wave/file.comp') and when linked into plugins-source linked into sfdx.

### DIFF
--- a/src/registry/registry.json
+++ b/src/registry/registry.json
@@ -1047,6 +1047,16 @@
         "adapter": "matchingContentFile"
       }
     },
+    "wavecomponent": {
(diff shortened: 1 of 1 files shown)


smithgp (approved):
Well, I think looks good 😄

PR #379, opened 2021-07-13
fix: resolveComponent no longer tries to resolve ignored dir paths

shetzel:
### What does this PR do?
Changes MetadataResolver.resolveComponent() to check if all paths are forceignored (including directories) rather than just metadata files before resolution so that errors aren't thrown in certain cases (see below).

This PR is courtesy of @jayree.  Thank you!

@W-9543674@

### Functionality Before
With `.forceignore` contents that include an entry of `**/test1`, which should ignore test1 directories the following command would cause an error:
`sfdx force:source:retrieve -p force-app`

ERROR running force:source:retrieve:  Metadata xml file unpackaged/experiences/test1.site-meta.xml is forceignored but is required for unpackaged/experiences/test1

### Functionality After
No error is thrown for the same command and the same `.forceignore` contents

### DIFF
--- a/src/resolve/metadataResolver.ts
+++ b/src/resolve/metadataResolver.ts
@@ -78,9 +78,11 @@ export class MetadataResolver {
       if (this.tree.isDirectory(fsPath)) {
         if (this.resolveDirectoryAsComponent(fsPath)) {
           const component = this.resolveComponent(fsPath, true);
-          if (!inclusiveFilter || inclusiveFilter.has(component)) {
-            components.push(component);
-            ignore.add(component.xml);
+          if (component) {
+            if (!inclusiveFilter || inclusiveFilter.has(component)) {
+              components.push(component);
+              ignore.add(component.xml);
+            }
           }
         } else {
           dirQueue.push(fsPath);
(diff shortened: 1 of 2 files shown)


src/resolve/metadataResolver.ts
```diff
@@ -78,9 +78,11 @@ export class MetadataResolver {
       if (this.tree.isDirectory(fsPath)) {
         if (this.resolveDirectoryAsComponent(fsPath)) {
           const component = this.resolveComponent(fsPath, true);
-          if (!inclusiveFilter || inclusiveFilter.has(component)) {
-            components.push(component);
-            ignore.add(component.xml);
+          if (component) {
+            if (!inclusiveFilter || inclusiveFilter.has(component)) {
```
WillieRuemmele, line 82: Can we combine the if statements to be ` if (component && (!inclusiveFilter || inclusiveFilter.has(component)))`

shetzel: sure

WillieRuemmele (approved):

WillieRuemmele:
passed QA, would not retrieve a forceignored experience bundle MDT

PR #380, opened 2021-07-14
refactor: rename taraj to mixedContentDirectory

violetyao:
### What does this PR do?
This PR renames the test constants for mixed content with directory as content types from "taraji" to "mixedContentDirectory". 

@W-9170998@

### Functionality Before

There is no change in functionality. 

### Functionality After

There is no change in functionality.


test/mock/registry/mockRegistry.ts
```diff
@@ -118,11 +118,11 @@ export const mockRegistryData = {
         recomposition: 'startEmpty',
       },
     },
-    tarajihenson: {
-      id: 'tarajihenson',
-      directoryName: 'tarajis',
+    mixedcontentdirectory: {
```
shetzel, line 121: It would really help people write unit tests using the mock registry if there were comments (like above) that matched these types to real world examples.

Also, a general comment at the top explaining what it is and why you'd use it rather than the actual registry.json.

shetzel (changes requested):

test/mock/registry/mockRegistry.ts
AnanyaJha: Great point, let's add a short comment above this entry similar to [what was done here ](https://github.com/forcedotcom/source-deploy-retrieve/blob/e0d89d300993a0972a90420a88d4b6e3a4b6e475/test/mock/registry/mockRegistry.ts#L85) @violetyao 

We can file a separate story to add the general comments about the mockregistry file vs registry.json

AnanyaJha (approved):

test/mock/registry/mockRegistry.ts
violetyao: Great point, thank you both for commenting! I've added a short comment & a real word example for mixedContentDirectory based on `mixedContentDirectoryConstants.ts`, usages of `mockRegistryData.types.mixedcontentdirectory`, and the [Metadata type doc](https://developer.salesforce.com/docs/atlas.en-us.api_meta.meta/api_meta/meta_types_list.htm). Let me know what you think! I've created a separate story for comments about `mockRegistry.ts` and the actual registry at W-9620029.

shetzel: Thank you Violet!

shetzel (approved):

PR #381, opened 2021-07-15
fix: add support for sourceApiVersion

shetzel:
### What does this PR do?
Adds the ability to set the manifest version from a ComponentSet.  This is necessary to support source that is in a format specific to an API version.  Also overloads the MetadataConverter.convert() method to take a ComponentSet or list of Components.

 @W-9604681@

### Functionality Before
It was not possible to change the version within the generated manifest for a deploy or retrieve

### Functionality After
The `sourceApiVersion` within a sfdx-project.json can be used to change the generated manifest version.

### DIFF
--- a/src/client/metadataTransfer.ts
+++ b/src/client/metadataTransfer.ts
@@ -146,19 +146,13 @@ export abstract class MetadataTransfer<
       try {
         const source = cs || this.components || new ComponentSet();
         const converter = new MetadataConverter();
+        await converter.convert(source, target, {
+          type: 'directory',
+          outputDirectory: mdapiTempDir,
+        });
         if (target === 'source') {
-          await converter.convert(source.getSourceComponents().toArray(), target, {
-            type: 'directory',
-            outputDirectory: mdapiTempDir,
-          });
-          // for source convert the package.xml isn't included, we'll write that separately
-
+          // for source convert the package.xml isn't included so write it separately
           fs.writeFileSync(join(mdapiTempDir, 'package.xml'), source.getPackageXml());
-        } else {
-          await converter.convert(source.getSourceComponents().toArray(), target, {
-            type: 'directory',
-            outputDirectory: mdapiTempDir,
-          });
         }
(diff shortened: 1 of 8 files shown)


WillieRuemmele (approved):

src/client/metadataTransfer.ts
```diff
@@ -171,7 +165,7 @@ export abstract class MetadataTransfer<
       this.usernameOrConnection = await Connection.create({
         authInfo: await AuthInfo.create({ username: this.usernameOrConnection }),
       });
-      if (this.apiVersion) {
+      if (this.apiVersion && this.apiVersion !== this.usernameOrConnection.version) {
```
shetzel, line 168: Only set it and report that it's being overridden when the version has been set and is different.

src/collections/componentSet.ts
```diff
@@ -246,7 +254,7 @@ export class ComponentSet extends LazyCollection<MetadataComponent> {
     return {
       Package: {
         types: typeMembers,
-        version: this.apiVersion,
+        version: this.sourceApiVersion || this.apiVersion,
```
shetzel, line 257: use the `sourceApiVersion` before defaulting to the API version being used.

src/convert/metadataConverter.ts
```diff
@@ -49,10 +49,36 @@ export class MetadataConverter {
...
+   *
+   * @param componentSet ComponentSet to convert
+   * @param targetFormat Format to convert the component files to
+   * @param output Configuration for outputting the converted files
+   */
+  public async convert(
+    componentSet: ComponentSet,
+    targetFormat: SfdxFileFormat,
+    output: ConvertOutputConfig
+  ): Promise<ConvertResult>;
```
shetzel, line 65: This overloads the convert function to accept a ComponentSet or list of SourceComponents.  This is a non-breaking way to support both, and passing the ComponentSet is required if you want to maintain the same sourceApiVersion and apiVersion for the convert.

PR #383, opened 2021-07-15
chore: enable EmailServicesFunction

WillieRuemmele:
### What does this PR do?
This PR goes against the [handbook](https://salesforce.quip.com/F3q6AmNSquvE#ZJKACAn1eZP) but enabling this type for a retrieve which would have required adding a new Transfomer/adapter, which would also go against the [handbook](https://salesforce.quip.com/F3q6AmNSquvE#ZJKACABWVSO)

this approach seemed a little better

all credit goes to @jayree

@W-9543539@

### Functionality Before
`EmailServicesFunction` wasn't able to be retrieved
<insert gif and/or summary>

### Functionality After
`EmailServicesFunction` able to be retrieved

<insert gif and/or summary>

### DIFF
--- a/scripts/update-registry/typeOverride.json
+++ b/scripts/update-registry/typeOverride.json
@@ -1,7 +1,4 @@
 {
-  "emailservicesfunction": {
-    "suffix": null
-  },
   "emailtemplatefolder": {
     "id": "emailfolder",
     "name": "EmailFolder",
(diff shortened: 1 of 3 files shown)


xyc:
@WillieRuemmele do you want to update / add a section in the handbook to explain this change?

Currently it says:
> For instance, the file suffix for EmailServicesFunction is xml which causes problems - that’s considered a reserved suffix. We add an entry in the type override file to make sure the suffix is nulled out when the script attempts to update the entry.

shetzel (approved):

shetzel:
I tested with ebikes_lwc repo and a few others that have EmailServicesFunction.  Deployed/retrieved in various ways and it all looks good.

PR #393, opened 2021-07-22
fix: add missing types from toolbelt to the best of my ability

WillieRuemmele:
### What does this PR do?
Adds missing metadata types to the registry that toolbelt supported

@W-9604298@

### Functionality Before
SDR would throw an error `Missing metadata type definition in registry for id 'dynamictrigger'`

### Functionality After
SDR should be able to work with the type


shetzel (approved):

shetzel:
Tested with ebikes_lwc, dreamhouse, and another large project.  Deploy, retrieve and convert.  All good.

PR #396, opened 2021-07-22
chore: add comments for mockRegistry vs actual registry

violetyao:
### What does this PR do?
We mentioned in #380 that it would be helpful for users if there is a comment at the top of `mockRegistry.ts` explaining what it is and why they would use it rather than the actual registry.json.

@W-9620029@, #380 

### Functionality Before

There is no change in functionality.
### Functionality After

There is no change in functionality.


AnanyaJha (approved):

AnanyaJha:
@shetzel Let us know if you were looking for anything else!

shetzel (approved):
Thank you!

PR #412, opened 2021-08-02
chore: invalid empty bundle types filtered from result

WillieRuemmele:
### What does this PR do?

if a there was an empty top level directory for a bundle type, I'll use LWC as an example, it would cause deploy errors. I filter those out during the bundle adapter's `populate` method

so a `force:source:deploy -m LightningComponentBundle`
on a dir that looks like 
```   * lwc/
   * ├── validLWC/
   * |   ├── myFoo.js
   * |   ├── myFooStyle.css
   * |   ├── myFoo.html
   * |   ├── myFoo.js-meta.xml
   * ├── invalidLWC/
```

will only result in `validLWC` being deployed, and no errors

#https://github.com/forcedotcom/cli/issues/1099, @W-9692466@

### Functionality Before
error on the server `An object 'withoutContent' of type LightningComponentBundle was named in package.xml, but was not found in zipped directory`
<insert gif and/or summary>

### Functionality After
deploys correctly
<insert gif and/or summary>

### DIFF
--- a/src/resolve/adapters/bundleSourceAdapter.ts
+++ b/src/resolve/adapters/bundleSourceAdapter.ts
@@ -47,15 +47,10 @@ export class BundleSourceAdapter extends MixedContentSourceAdapter {
    * @protected
    */
   protected populate(trigger: SourcePath, component?: SourceComponent): SourceComponent {
-    if (this.tree.isDirectory(trigger) && this.tree.readDirectory(trigger)?.length) {
-      // if we're populating from a populated directory (validLWC)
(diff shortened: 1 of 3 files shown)


WillieRuemmele (commented):
LGTM, thanks @shetzel

shetzel (approved):
Verified unit test fails without the code fix and passes with it.
Verified the fix works using the ebikes_lwc project, with and without an empty LWC directory.
Code looks good with nice jsdoc comments branched from develop.

PR #413, opened 2021-08-03
fix: forceIgnore does not work for `SFDX: Retrieve Source in Manifest from Org` in vscode extension

violetyao:
### What does this PR do?
1) When users use the `SFDX: Retrieve Source in Manifest from Org` command in vscode extension, 1) files in `.forceIgnore` are not printed in `channelService` (which is the expected behavior) 2) however, files in `.forceIgnore` are still retrieved and they overwrite the local version (which is not the expected behavior). 

This PR fixes the issue where forceIgnore does not work for `SFDX: Retrieve Source in Manifest from Org` in vscode extension, by keeping track of a list of forceIgnoredPaths explored in resolver and skip those forceIgnoredPaths during writing. 

2) This PR also fixes  @W-9682706@ where forceIgnored matchingContent files are retrieved to `force-app/main/default/classes/ApexClassExample.cls`, while their original paths are `force-app/main/default/classes/subfolder/ApexClassExample.cls`.

https://github.com/forcedotcom/salesforcedx-vscode/issues/3236, @W-9293630@, @W-9682706@

### Functionality Before

.forceIgnore did not work for retrieving a componentSet constructed from `ComponentSet.fromManifest`.

### Functionality After

.forceIgnore works for retrieving a componentSet constructed from `ComponentSet.fromManifest`.


src/convert/types.ts
```diff
@@ -64,6 +64,7 @@ export type MergeConfig = {
    * Location to store components that aren't merged.
    */
   defaultDirectory: SourcePath;
+  forceIgnoredPaths?: string[];
```
xyc, line 67: Can we use a [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) for ignoredPaths? `Array.prototype.includes` is an O(n) operation while `Set.prototype.has` is amortized constant time.

violetyao: Will do, thanks Xiaoyi!

src/convert/streams.ts
```diff
@@ -150,9 +151,22 @@ export class StandardWriter extends ComponentWriter {
...
+          let fullDest = isAbsolute(info.output)
             ? info.output
             : join(this.rootDestination, info.output);
+          if (!fs.fileExistsSync(fullDest)) {
+            for (const ignoredPath of this.forceIgnoredPaths) {
+              if (
+                dirname(ignoredPath).includes(dirname(fullDest)) &&
+                basename(ignoredPath).includes(basename(fullDest))
+              ) {
+                fullDest = ignoredPath;
```
jag-sfdc, line 163: Can we just return here? I think this code can be simplified a bit.

violetyao: Good point! Updated. Thanks Jag!

jag-sfdc (approved):

PR #418, opened 2021-08-09
fix: getComponent won't throw when parent is ignored

WillieRuemmele:
### What does this PR do?
the `decomposedSourceAdapter` was calling `baseSourceAdapter.getComponents` which was throwing an error when a parent xml file wasn't found or was .forceignore'd

this wasn't working for decomposed types because the parent isn't required... and so it was throwing when it shouldn't - blocking functionality. 

I overrode  the `getComponents` in the `decomposedSourceAdapter` adapter so that it wouldn't throw an error in that scenario

https://github.com/forcedotcom/cli/issues/1110, https://github.com/forcedotcom/cli/issues/1106, @W-9720135@

### Functionality Before
```
Metadata xml file /Users/william.ruemmele/projects/scratches/SFDXCLI-Issues/CLIIssues/force-app/main/default/objects/hed__Term__c/hed__Term__c.object-meta.xml is forceignored but is required for /Users/william.ruemmele/projects/scratches/SFDXCLI-Issues/CLIIssues/force-app/main/default/objects/hed__Term__c/fields/Term_Title__c.field-meta.xml
```
<insert gif and/or summary>

### Functionality After
`Source was successfully converted to Metadata API format`
<insert gif and/or summary>

### DIFF
--- a/src/resolve/adapters/baseSourceAdapter.ts
+++ b/src/resolve/adapters/baseSourceAdapter.ts
@@ -54,7 +54,7 @@ export abstract class BaseSourceAdapter implements SourceAdapter {
 
     let component: SourceComponent;
     if (rootMetadata) {
-      const componentName = this.type.inFolder
+      const componentName = this.type.folderType
         ? `${parentName(rootMetadata.path)}/${rootMetadata.fullName}`
         : rootMetadata.fullName;
       component = new SourceComponent(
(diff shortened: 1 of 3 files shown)


src/resolve/adapters/baseSourceAdapter.ts
```diff
@@ -54,7 +54,7 @@ export abstract class BaseSourceAdapter implements SourceAdapter {
 
     let component: SourceComponent;
     if (rootMetadata) {
-      const componentName = this.type.inFolder
+      const componentName = this.type.folderType
```
WillieRuemmele, line 57: `inFolder` was deprecated, `folderType` was the suggested replacement

shetzel (approved):
Tested several converts/deploys/retrieves pointing to CustomObject children, child dirs, and object dirs with the object meta file ignored, not present, and present.  All looks good.

PR #424, opened 2021-08-16
fix: xml nodes might not be arrays

mshanemc:
### What does this PR do?
don't call `push` on something that isn't guaranteed to be an array.
reproducer here: https://github.com/mshanemc/github1119 
  
https://github.com/forcedotcom/cli/issues/1119, @W-9747902@

### Functionality Before

<insert gif and/or summary>

### Functionality After

<insert gif and/or summary>


WillieRuemmele (approved):
Verified the changes in the linked repo working as intended

PR #425, opened 2021-08-17
refactor: update SDR to be a mono-repo

violetyao:
### What does this PR do?
As part of an effort to make tests involving SourceComponents better (@W-9662066@), Max copied several files from SDR to the extensions. These files now live in both `salesforcedx-vscode-core/test/vscode-integration/mock/registry` and in `sdr/test/mock/resgistry`. Instead of having the files in both places, they should exist in one place in their own package that both the extensions and SDR can import. 

We discussed three approaches to create the package:

1) create a separate repo: source-deploy-retrieve-test-utils, the con is that it would bring extra release work
2) move `mock/registry` from `/test` to `src/`, the con is that it would expose test constants to those who use SDR, and those who only want to use test constants would have to add SDR as a dependency
3) update SDR to be a mono repo, containing packages source-deploy-retrieve and source-deploy-retrieve-test-utils

We decided to proceed with the third approach. This PR updates SDR to a mono repo, a subsequent PR would add the`source-deploy-retrieve-test-utils` packag, which contains test constants under `mock/registry`.

https://github.com/forcedotcom/salesforcedx-vscode/pull/3481, @W-9682367@

### Functionality Before

SDR was not a mono repo. 

### Functionality After

SDR is updated to a mono repo.


jag-sfdc:
Based on discussions with CLI, we want to move SDR to their release style and releasing multiple repos is fairly easy. So, the design 1 proposed in the description above will be the preferred option. But we can get to it at a later point. Closing this PR for now.

PR #426, opened 2021-08-17
fix: throw an error for unexpected child types

shetzel:
### What does this PR do?
When resolving source and an unexpected child type (e.g., ApexClass) is found within a strictFolderType (e.g., CustomObject) an error will be thrown.

github issue: https://github.com/forcedotcom/cli/issues/1106
@W-9764876@

### Functionality Before
```
TypeError: Cannot set property 'content' of undefined
at DecomposedSourceAdapter.populate (source-deploy-retrieve/lib/src/resolve/adapters/decomposedSourceAdapter.js:77:35)
```

### Functionality After
Better error with more information about the cause of the problem.  E.g.,
```
Unexpected child metadata [/path/to/project/force-app/main/default/objects/Case/layouts/Case-Case Layout.layout-meta.xml] found for parent type [CustomObject] 
```


WillieRuemmele (approved):
with a project like `objects/<my object>/classes/myclass.cls` it would throw the error when

- convert
- deploy
- retrieve

tested with layouts, classes, pages under custom objects, and under other metadata types

PR #427, opened 2021-08-17
fix: force fullName to CustomLabels

mdonnalley:
### What does this PR do?

Adds `ignoreParsedFullName` option to registry. When `true` this will ignore the name that was parsed from the file path and use the metadata type instead. Doing so fixes deploy and retrieve for users who have CustomLabels files _not_ named `CustomLabels.labels-meta.xml`, e.g. `MyLabels.labels-meta.xml`

https://github.com/forcedotcom/cli/issues/1133, 
@W-9720198@


src/resolve/sourceComponent.ts
```diff
@@ -218,6 +218,9 @@ export class SourceComponent implements MetadataComponent {
   }
 
   get fullName(): string {
+    if (this.type.ignoreTypeName) {
```
WillieRuemmele, line 221: NIT: the naming makes this line confusing, if `ignoreTypeName` is true, we return `this.type.name` which I wouldn't expect to happen if I set `ignoreTypeName=true`. From the JSDoc, what if we change `ignoreTypeName` to `ignoreParsedFileName`

But I confirmed the change works, 

- deploy `CustomLabels.labels-meta.xml`
-  deploys `Labels.labels-meta.xml`
- deploys `force-app/main/default/myLabels/Labels.labels-meta.xml`
- retrieves into `myLabels/Labels...`
- retrieves into `labels/CustomLabels..`

WillieRuemmele (approved):

src/resolve/sourceComponent.ts
mshanemc: I test this on MPD deploy-retrieve scenarios and did `source:convert` with it, too.

PR #428, opened 2021-08-17
fix: support toolbelt suffixes

mshanemc:
### What does this PR do?
toolbelt had two differences with SDR on metadata suffixes.  This changes adds an additional suffix key pointing to the correct metadata type for index (big object indexes, child of customObject) and customwebpagelink.

For top-level type (customwebpagelink, not indexes) this also adds a new registry type property called `legacySuffix`.  If that exists, the destination xml location does a `replace` to swap out the suffix (the existing code was just `slice`ing off `meta-xml` which would preserve incorrect suffixes even when SDR knew the proper type.  That should provide deployable, corrected output for `convert` and `deploy` commands.

This should have no effect on any other metadata types where there is no `legacySuffix` provided.
 
https://github.com/forcedotcom/cli/issues/1141 , @W-9771788@ @W-9762131@

AC: files should deploy with either or both suffix/types
files with the toolbelt type should retrieve/merge into the existing file
newly retrieved files should only use the new, correct SDR suffixes

please test source:convert with this


WillieRuemmele (commented):
CustomPageWebLink:

- image
- is able to deploy both suffixes (above)
- CANNOT deploy `.custompageweblink` suffix without a `.weblink` file -> `"An object 'test' of type CustomPageWebLink was named in package.xml, but was not found in zipped directory"`
- retrieve will create a `.weblink` file 
- retrieve won't override a `.custompageweblik` to `.weblink`

Index:

- deploy works with test repo on this SDR branch, fails on develop
- source:convert works
- source:convert -> mdapi:convert will throw `Unexpected file found in package directory: /Users/william.ruemmele/projects/scratches/github1141-big-object-index/force-app/main/default/objects/Test__b/indexes/TestIndex.index-meta.xml`
- change that file it complains about above to use `.indexe` and it works
- retrieve works
- SDR will retrieve and write as `indexe` and `index`

WillieRuemmele:
Now able to deploy singular `.custompageweblink` file

WillieRuemmele (approved):

PR #429, opened 2021-08-18
Wr/content type of undefined

WillieRuemmele:
### What does this PR do?
When a static resource was missing it's `.resource` file it would throw a `Cannot read property 'contentType' of undefined` error that wasn't helpful at all

this catches that scenario, and adds some info to the message to allow the user to more easily debug and fix their source

#https://github.com/forcedotcom/cli/issues/1137, @W-9766036@

### Functionality Before

`Cannot read property 'contentType' of undefined`

### Functionality After
`A StaticResource must have an associated .resource file, missing staticresources/srfolder.resource-meta.xml`


src/convert/transformers/staticResourceMetadataTransformer.ts
```diff
@@ -141,7 +141,15 @@ export class StaticResourceMetadataTransformer extends BaseMetadataTransformer {
   }
 
   private async getContentType(component: SourceComponent): Promise<string> {
-    return ((await component.parseXml()).StaticResource as JsonMap).contentType as string;
+    try {
+      return ((await component.parseXml()).StaticResource as JsonMap).contentType as string;
```
mshanemc, line 145: It looks like you could create or pass in some type for this since the signature is
```
  public async parseXml<T = JsonMap>(): Promise<T> {
```

like component.parseXml<{StaticResource: {contentType: string;};}>()

mshanemc (approved):
this produces a nice error message when the resource-meta.xml is missing on a folder static resource (zip)

PR #433, opened 2021-08-23
sm/territory2

mshanemc:
### What does this PR do?
makes territory2 objects deploy correctly

https://github.com/forcedotcom/cli/issues/1147 @W-9789567@

### Functionality Before

created rules/territories in top-level folder, wouldn't deploy

### Functionality After
rules/territoris are children inside the territoryModel

### DIFF
--- a/src/resolve/adapters/baseSourceAdapter.ts
+++ b/src/resolve/adapters/baseSourceAdapter.ts
@@ -54,12 +54,9 @@ export abstract class BaseSourceAdapter implements SourceAdapter {
 
     let component: SourceComponent;
     if (rootMetadata) {
-      const componentName = this.type.folderType
-        ? `${parentName(rootMetadata.path)}/${rootMetadata.fullName}`
-        : rootMetadata.fullName;
       component = new SourceComponent(
         {
-          name: componentName,
+          name: this.calculateName(rootMetadata),
           type: this.type,
           xml: rootMetadata.path,
         },
@@ -144,6 +141,24 @@ export abstract class BaseSourceAdapter implements SourceAdapter {
     if (match && !match[1].includes('.') && parts.length > 1) {
       return { fullName: match[1], suffix: undefined, path: fsPath };
     }
+  }
+
+  private calculateName(rootMetadata: MetadataXml): string {
+    if (!this.type.folderType) {
+      return rootMetadata.fullName;
+    }
+    const grandparentType = this.registry.getTypeByName(this.type.folderType);
+    if (grandparentType.folderType && grandparentType.folderType !== this.type.id) {
+      const splits = rootMetadata.path.split(sep);
+      return `${splits[splits.indexOf(grandparentType.directoryName) + 1]}.${
+        rootMetadata.fullName
+      }`;
+    }
+    if (grandparentType.folderType === this.type.id) {
+      return rootMetadata.fullName;
+    }
+
+    return `${parentName(rootMetadata.path)}/${rootMetadata.fullName}`;
   }
 
   /**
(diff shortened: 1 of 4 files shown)


src/resolve/adapters/baseSourceAdapter.ts
```diff
@@ -146,6 +146,24 @@ export abstract class BaseSourceAdapter implements SourceAdapter {
...
+      return `${splits[splits.indexOf(grandparentType.directoryName) + 1]}.${
+        rootMetadata.fullName
+      }`;
+    }
+    if (grandparentType.folderType === this.type.id) {
+      return rootMetadata.fullName;
+    }
+
+    return `${parentName(rootMetadata.path)}/${rootMetadata.fullName}`;
+  }
```
shetzel, line 165: I think this could use some debug or at least a comment for each return providing an example.  Most of this logic is for folder types so maybe it would improve readability to do something like below this instead?

In `getComponent()` just before creating a new SourceComponent:
```javascript
let fullName = rootMetadata.fullName;
let parentType;
if (this.type.folderType) {
  parentType = this.registry.getTypeByName(this.type.folderType);
  fullName = getFolderFullName(rootMetadata,  parentType);
}
```

Then in getFolderFullName it would be mostly the same as calculateName just only about folders
```javascript
private getFolderFullName(rootMetadata, parentType) { ... }
```

src/resolve/adapters/baseSourceAdapter.ts
```diff
@@ -146,6 +146,24 @@ export abstract class BaseSourceAdapter implements SourceAdapter {
     }
   }
 
+  private calculateName(rootMetadata: MetadataXml): string {
+    if (!this.type.folderType) {
+      return rootMetadata.fullName;
+    }
+    const grandparentType = this.registry.getTypeByName(this.type.folderType);
+    if (grandparentType.folderType && grandparentType.folderType !== this.type.id) {
+      const splits = rootMetadata.path.split(sep);
```
shetzel, line 155: Using path sep can be tricky because if the path is from a server response it will necessarily be a forward slash, even on Windows OS.  If this is used for a retrieve on Windows I'd expect it to not work as expected.  In order for it to work we first have to normalize the path, then split.

src/resolve/sourceComponent.ts
```diff
@@ -129,6 +117,29 @@ export class SourceComponent implements MetadataComponent {
...
+      return trimUntil(fsPath, directoryName);
+    }
+    // legacy version of folderType
+    if (inFolder) {
+      return join(directoryName, this.fullName.split('/')[0], basename(fsPath));
+    }
+    if (folderType) {
+      // types like Territory2Model have child types inside them.  We have to preserve those folder structures
+      if (this.parentType?.folderType && this.parentType?.folderType !== this.type.id) {
+        const fsPathSplits = fsPath.split(sep);
```
shetzel, line 135: I'll echo the same comment from above about using `path.sep` to split file paths.  For this to work we have to normalize the path first unless we are absolutely sure this path can only come from the file system of the OS.

src/resolve/adapters/baseSourceAdapter.ts
mshanemc: I added some explanatory comments with examples.

That's an elegant solution for the current situation.  I think we'll  have a few more flavors coming (sharing*Rule) where the names also have wrinkles.  

I was wanting one place to keep all that stuff.

mshanemc: would it be ok to add our normal logger stuff in here?  What if there were like 1000 components?  BSA doesn't have a way to know the size...it would need to be told that somewhere.

shetzel: Maybe debug isn't a good option then.

mshanemc: it's hard to do in SDR...you'd be passing it down multiple levels to get to some of stream transformer stuff.

PR #436, opened 2021-08-24
Orb

mshanemc:
What does this PR do?
SDR should use the orb and build steps similar sfdx-core does (release off of every commit to main, don't sign).
no publish approval required
And also run the plugin-source NUTs [...has some GC/memory errors consistent with plugin-source itself]. That's why the external NUT is not required for publish yet
retain existing slack notifications to cli-notifications channel (does not include version number)
verify windows tests work with our process
update documentation in the SDR repo to match the new process for contributing, developing, and publishing.
bumps package.json one minor version to "QA" a release when merged into main.

What issues does this PR fix or reference?
@W-9789170@

PR #437, opened 2021-08-25
orb-from-main

mshanemc:
### What does this PR do?

[redo of https://github.com/forcedotcom/source-deploy-retrieve/pull/436 to get correct commit history]

SDR should use the orb and build steps similar sfdx-core does (release off of every commit to main, don't sign).
no publish approval required
And also run the plugin-source NUTs [...has some GC/memory errors consistent with plugin-source itself]. That's why the external NUT is not required for publish yet
retain existing slack notifications to cli-notifications channel (does not include version number)
verify windows tests work with our process
update documentation in the SDR repo to match the new process for contributing, developing, and publishing.
bumps package.json one minor version to "QA" a release when merged into main.

@W-9789170@


mshanemc:
still picking up wrong commits

PR #443, opened 2021-08-26
Wr/aura under object main

WillieRuemmele:
### What does this PR do?
When an Aura's `.cmp-meta.xml` file was under a custom object it would throw a `Cannot read property 'id' of undefined` error. handle it by refactoring the `ensureValidChildren` method to the `SourceComponent` class and call in that scenario

#https://github.com/forcedotcom/cli/issues/1148, @W-9800691@

### Functionality Before
`Cannot read property 'id' of undefined` 

### Functionality After

`Unexpected child metadata [/Users/william.ruemmele/projects/scratches/dreamhouse-lwc/force-app/main/default/objects/Broker__c/pageTemplate_2_7_3.cmp-meta.xml] found for parent type [CustomObject]`

### DIFF
--- a/src/resolve/sourceComponent.ts
+++ b/src/resolve/sourceComponent.ts
@@ -93,16 +93,21 @@ export class SourceComponent implements MetadataComponent {
     }
     return {} as T;
   }
-  // Ensures that the children of the provided SourceComponent are valid child
-  // types before adding them to the returned ComponentSet. Invalid child types
-  // can occur when projects are structured in an atypical way such as having
-  // ApexClasses or Layouts within a CustomObject folder.
+
+  /**
+   * Ensures that the children of the provided SourceComponent are valid child
+   * types before adding them to the returned ComponentSet. Invalid child types
+   * can occur when projects are structured in an atypical way such as having
+   * ApexClasses or Layouts within a CustomObject folder.
+   * @param compSet a component set to add children to
+   * @param registry the metadata registry access
+   */
   public ensureValidChildren(
     compSet?: ComponentSet,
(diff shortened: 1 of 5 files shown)


src/resolve/sourceComponent.ts
```diff
@@ -76,12 +76,28 @@ export class SourceComponent implements MetadataComponent {
     }
     return sources;
   }
-
+  /**
+   * Ensures that the children of SourceComponent are valid child
+   * types. Invalid child types can occur when projects are structured in an atypical way such as having
+   * ApexClasses or Layouts within a CustomObject folder.
```
shetzel, line 82: Update the comment a bit to more accurately reflect this method.

src/resolve/sourceComponent.ts
```diff
@@ -14,7 +14,9 @@ import { DEFAULT_PACKAGE_ROOT_SFDX } from '../common';
 import { get, getString, JsonMap } from '@salesforce/ts-types';
 import { SfdxFileFormat } from '../convert';
 import { trimUntil } from '../utils/path';
-import { MetadataType } from '../registry';
+import { MetadataType, RegistryAccess } from '../registry';
+import { ComponentSet } from '../collections';
```
shetzel, line 18: Remove ComponentSet and any others that are no longer necessary

shetzel (changes requested):

shetzel (approved):

shetzel:
I tested a few deploys, retrieves and converts with an aura meta file in a custom object dir and see the error message.  Removing the file I could deploy, retrieve, and convert successfully.

PR #444, opened 2021-08-31
Sh/exclude custom field translations

shetzel:
### What does this PR do?
Adds the ability to prevent metadata type entries from being written to the manifest.  These are special cases where the type must be defined in the registry so it can be identified and possibly decomposed/recomposed, but they are not supported by the Metadata API.

Example type: CustomFieldTranslation

@W-9831738@

### Functionality Before
The type entry was added to the manifest and the deploy or retrieve request would fail with an error like this:
`Unknown type name 'CustomFieldTranslation' specified in package.xml`

### Functionality After
Successful deploy/retrieve because the type is not added to the manifest.

### DIFF
--- a/src/collections/componentSet.ts
+++ b/src/collections/componentSet.ts
@@ -289,7 +289,9 @@ export class ComponentSet extends LazyCollection<MetadataComponent> {
       const componentMap = components.get(key);
       for (const comp of componentMap.values()) {
         for (const child of comp.getChildren()) {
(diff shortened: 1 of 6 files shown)


WillieRuemmele (approved):
Nice, super generic and easy to add another type like this!

WillieRuemmele:
verified it would write manifests with only `CustomObjectTranslation` when converting/deploying/retrieving 

and I removed the `**/objectTranslations/**` entry from the `.forceignore`

PR #447, opened 2021-09-01
fix: force RestrictionRule to a strict directory

shetzel:
### What does this PR do?
Updates the registry to define a strictDirectoryName for the RestrictionRule type since the suffix ("rule") is also used by the ModerationRule type.

 https://github.com/forcedotcom/cli/issues/1158
@W-9833678@

### Functionality Before
Source is converted to metadata API format with a moderation folder instead of restrictionRules. The package.xml file also defines the metadata type as ModerationRule instead of RestrictionRule.

### Functionality After
Source is converted to metadata API format with a restrictionRules folder and the package.xml should reference the RestrictionRule metadata type.

### DIFF
--- a/src/registry/registry.json
+++ b/src/registry/registry.json
@@ -2069,7 +2069,7 @@
       "name": "RestrictionRule",
       "suffix": "rule",
       "directoryName": "restrictionRules",
-      "strictDirectoryName": false
+      "strictDirectoryName": true
     },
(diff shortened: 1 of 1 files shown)


WillieRuemmele (approved):
Confirmed converted `RestrictionRules` correctly with the issue's repo

PR #448, opened 2021-09-09
fix: fix node 16 error parsing issue

WillieRuemmele:
### What does this PR do?
fixes an Error message parsing issue with differences between `node 14` and `node 16`

### Functionality Before
`Cannot read properties of undefined (reading 'contentType')`

### Functionality After
`A StaticResource must have an associated .resource file, missing staticresources/a.resource-meta.xml`


mshanemc (approved):
this is much improved to not swallow the other types of errors.  now getContentType can throw.

PR #450, opened 2021-09-13
Wr/destructive changes

WillieRuemmele:
### What does this PR do?
adds pre/post destructive change support to `ComponentSet.fromManifest`

@W-9892375@

### Functionality Before
destructive component sets were limited to a single type (pre/post)
<insert gif and/or summary>

### Functionality After
you can now specify which type of deletion per `SourceComponent`
<insert gif and/or summary>

### DIFF
--- a/src/collections/componentSet.ts
+++ b/src/collections/componentSet.ts
@@ -13,19 +13,19 @@ import {
 import { XML_DECL, XML_NS_KEY, XML_NS_URL } from '../common';
 import { ComponentSetError } from '../errors';
 import {
-  MetadataResolver,
+  ComponentLike,
   ManifestResolver,
+  MetadataComponent,
+  MetadataResolver,
   SourceComponent,
   TreeContainer,
-  MetadataComponent,
-  ComponentLike,
 } from '../resolve';
 import {
-  PackageTypeMembers,
-  FromManifestOptions,
-  PackageManifestObject,
-  FromSourceOptions,
   DestructiveChangesType,
+  FromManifestOptions,
+  FromSourceOptions,
+  PackageManifestObject,
+  PackageTypeMembers,
 } from './types';
 import { LazyCollection } from './lazyCollection';
 import { j2xParser } from 'fast-xml-parser';
@@ -64,7 +64,10 @@ export class ComponentSet extends LazyCollection<MetadataComponent> {
   private components = new Map<string, Map<string, SourceComponent>>();
 
   // internal component maps used by this.getObject() when building manifests.
-  private destructiveComponents = new Map<string, Map<string, SourceComponent>>();
+  private destructiveComponents = {
+    [DestructiveChangesType.PRE]: new Map<string, Map<string, SourceComponent>>(),
+    [DestructiveChangesType.POST]: new Map<string, Map<string, SourceComponent>>(),
+  };
   private manifestComponents = new Map<string, Map<string, SourceComponent>>();
 
   private destructiveChangesType = DestructiveChangesType.POST;
@@ -80,7 +83,7 @@ export class ComponentSet extends LazyCollection<MetadataComponent> {
       if (component instanceof SourceComponent) {
         asDeletion = component.isMarkedForDelete();
       }
-      this.add(component, asDeletion);
+      this.add(component, asDeletion, this.destructiveChangesType);
     }
   }
 
@@ -254,14 +257,17 @@ export class ComponentSet extends LazyCollection<MetadataComponent> {
    *
    * @returns Object representation of a package manifest
    */
-  public getObject(forDestructiveChanges = false): PackageManifestObject {
+  public getObject(
+    forDestructiveChanges = false,
+    destructiveType: DestructiveChangesType = this.destructiveChangesType
+  ): PackageManifestObject {
     // If this ComponentSet has components marked for delete, we need to
     // only include those components in a destructiveChanges.xml and
     // all other components in the regular manifest.
     let components = this.components;
     if (this.hasDeletes) {
       if (forDestructiveChanges) {
-        components = this.destructiveComponents;
+        components = this.destructiveComponents[destructiveType];
       } else {
         components = this.manifestComponents;
       }
@@ -327,13 +333,17 @@ export class ComponentSet extends LazyCollection<MetadataComponent> {
    * @param indentation Number of spaces to indent lines by.
    * @param forDestructiveChanges Whether to build a manifest for destructive changes.
    */
-  public getPackageXml(indentation = 4, forDestructiveChanges = false): string {
+  public getPackageXml(
+    indentation = 4,
+    forDestructiveChanges = false,
+    destructiveType: DestructiveChangesType = this.destructiveChangesType
+  ): string {
     const j2x = new j2xParser({
       format: true,
       indentBy: new Array(indentation + 1).join(' '),
       ignoreAttributes: false,
     });
-    const toParse = this.getObject(forDestructiveChanges);
+    const toParse = this.getObject(forDestructiveChanges, destructiveType);
     toParse.Package[XML_NS_KEY] = XML_NS_URL;
     return XML_DECL.concat(j2x.parse(toParse));
   }
@@ -360,7 +370,11 @@ export class ComponentSet extends LazyCollection<MetadataComponent> {
     >;
   }
 
-  public add(component: ComponentLike, asDeletion?: boolean): void {
+  public add(
+    component: ComponentLike,
+    asDeletion = false,
+    deletionType: DestructiveChangesType = this.destructiveChangesType
+  ): void {
     const key = this.simpleKey(component);
     if (!this.components.has(key)) {
       this.components.set(key, new Map<string, SourceComponent>());
@@ -373,10 +387,10 @@ export class ComponentSet extends LazyCollection<MetadataComponent> {
       if (asDeletion) {
         component.setMarkedForDelete(true);
         this.logger.debug(`Marking component for delete: ${component.fullName}`);
-        if (!this.destructiveComponents.has(key)) {
-          this.destructiveComponents.set(key, new Map<string, SourceComponent>());
+        if (!this.destructiveComponents[deletionType].has(key)) {
+          this.destructiveComponents[deletionType].set(key, new Map<string, SourceComponent>());
         }
-        this.destructiveComponents.get(key).set(this.sourceKey(component), component);
+        this.destructiveComponents[deletionType].get(key).set(this.sourceKey(component), component);
       } else {
         if (!this.manifestComponents.has(key)) {
           this.manifestComponents.set(key, new Map<string, SourceComponent>());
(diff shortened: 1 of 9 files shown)


src/resolve/sourceComponent.ts
```diff
@@ -131,8 +133,23 @@ export class SourceComponent implements MetadataComponent {
     return this.markedForDelete;
   }
 
-  public setMarkedForDelete(asDeletion: boolean): void {
-    this.markedForDelete = asDeletion;
+  public getDestructiveChangesType(): DestructiveChangesType {
+    return this.destructiveChangesType;
+  }
+
+  public setMarkedForDelete(destructiveChangeType?: DestructiveChangesType | boolean): void {
```
WillieRuemmele, line 140: as a setter, we need to be able to specify if it's a deletion (true/false) and what type of deletion`DestructiveChangesType` passing nothing will set it to `true` and a `post` delete

src/collections/componentSet.ts
```diff
@@ -360,7 +402,7 @@ export class ComponentSet extends LazyCollection<MetadataComponent> {
     >;
   }
 
-  public add(component: ComponentLike, asDeletion?: boolean): void {
+  public add(component: ComponentLike, deletionType?: DestructiveChangesType): void {
```
WillieRuemmele, line 405: replacing the `asDeletion` param is the breaking change, which can also be seen on other methods

src/collections/componentSet.ts
```diff
@@ -168,23 +171,62 @@ export class ComponentSet extends LazyCollection<MetadataComponent> {
...
+        resolveIncludeSet.add(component, deletionType);
       }
       const memberIsWildcard = component.fullName === ComponentSet.WILDCARD;
       if (!memberIsWildcard || options.forceAddWildcards || !options.resolveSourcePaths) {
-        result.add(component);
+        result.add(component, deletionType);
+      }
+    };
+
+    for (const component of manifest.components) {
```
WillieRuemmele, line 207: write the manifest comps. and then the destructive post (if present) and the destructive pre (if present)

test/convert/metadataConverter.test.ts
```diff
@@ -391,21 +404,21 @@ describe('MetadataConverter', () => {
       ]);
     });
 
-    it('should write destructive changes pre manifest when ComponentSet has deletes', async () => {
+    it('should write destructive changes pre manifest when ComponentSet has deletes marked for pre', async () => {
```
jag-sfdc, line 407: Nit pick. Can you rename the similar test for post in line 167?

src/collections/componentSet.ts
```diff
@@ -76,11 +79,11 @@ export class ComponentSet extends LazyCollection<MetadataComponent> {
     this.logger = Logger.childFromRoot(this.constructor.name);
 
     for (const component of components) {
-      let asDeletion = false;
+      let destructiveType = this.destructiveChangesType;
```
jeffb-sfdc, line 82: Since this is an "if-else" assignment, a turnery operation can be used here.
```
const destructiveType = (component instanceof SourceComponent)
  ? component.getDestructiveChangesType(),
  : this.destructiveChangesType;
```

src/collections/componentSet.ts
```diff
@@ -128,15 +131,15 @@ export class ComponentSet extends LazyCollection<MetadataComponent> {
 
     const resolver = new MetadataResolver(registry, tree);
     const set = new ComponentSet([], registry);
-    const buildComponents = (paths: string[], asDeletes: boolean): void => {
+    const buildComponents = (paths: string[], destructiveType?: DestructiveChangesType): void => {
       for (const path of paths) {
         for (const component of resolver.getComponentsFromPath(path, inclusiveFilter)) {
-          set.add(component, asDeletes);
+          set.add(component, destructiveType);
```
jeffb-sfdc, line 137: Just dbl-checking here, as I'm not clear on the code paths - since `destructiveType` is nullable, what happens when `set.add(component, null)` is called and then later used?

src/collections/componentSet.ts
```diff
@@ -254,14 +296,14 @@ export class ComponentSet extends LazyCollection<MetadataComponent> {
...
-  public getObject(forDestructiveChanges = false): PackageManifestObject {
+  public getObject(destructiveType?: DestructiveChangesType): PackageManifestObject {
     // If this ComponentSet has components marked for delete, we need to
     // only include those components in a destructiveChanges.xml and
     // all other components in the regular manifest.
     let components = this.components;
     if (this.hasDeletes) {
-      if (forDestructiveChanges) {
-        components = this.destructiveComponents;
+      if (destructiveType) {
```
jeffb-sfdc, line 305: A turnery operation can be used here

src/collections/componentSet.ts
WillieRuemmele: it should default to a `POST` i.e. generating a `destructiveChangesPost.xml`

src/collections/componentSet.ts
```diff
@@ -168,23 +171,62 @@ export class ComponentSet extends LazyCollection<MetadataComponent> {
...
     const options = (typeof input === 'object' ? input : {}) as Partial<FromManifestOptions>;
+    let manifestPre;
+    let manifestPost;
+    const hasManifestPre = !!options.destructivePre;
+    const hasManifestPost = !!options.destructivePost;
 
     const manifestResolver = new ManifestResolver(options.tree, options.registry);
     const manifest = await manifestResolver.resolve(manifestPath);
+    if (hasManifestPre) {
+      manifestPre = await manifestResolver.resolve(options.destructivePre);
```
mshanemc, line 182: suggest

```ts
const manifestPre = options.destructivePost ? await manifestResolver.resolve(options.destructivePre) : undefined;
```
then you could even refer to the truthiness of `manifestPre` further down

it's cleaner if everything can stay `const`

src/collections/componentSet.ts
```diff
@@ -370,19 +412,32 @@ export class ComponentSet extends LazyCollection<MetadataComponent> {
...
         if (!this.manifestComponents.has(key)) {
           this.manifestComponents.set(key, new Map<string, SourceComponent>());
         }
         this.manifestComponents.get(key).set(this.sourceKey(component), component);
       }
+
+      // something could try adding a component meant for deletion improperly, which would be marked as an addition
+      // specifically the ComponentSet.fromManifest with the `resolveSourcePaths` options which calls
+      // ComponentSet.fromSource, and adds everything as an addition
+      if (
```
mshanemc, line 433: oh, maybe I understand the flow logic now that I read down to here.  Maybe this could be a function that could run on the delete and non-delete path?

Or those paths could check for existence on its counterpart (deletion would delete it from manifestComponents and non-deletion would just not add it if it's already in deletions?)

src/collections/componentSet.ts
```diff
@@ -474,6 +529,24 @@ export class ComponentSet extends LazyCollection<MetadataComponent> {
     return this.destructiveChangesType;
   }
 
+  /**
+   * Will return the types of destructive changes in the component set
+   *
+   * @return DestructiveChangesType[]
+   */
+  public getTypesOfDestructiveChanges(): DestructiveChangesType[] {
```
mshanemc, line 537: `hasDeletes` is doing pretty much the same logic, and is checked in both the Converter and CS's getObject.

This could also return an empty array for "there are none" so there is only one.

As written, if you were to NOT check hasDeletes before asking `getTypesOfDestructiveChanges`, your CS would tell that both types are present (the else condition here).

suggestion: make an array, and conditionally push to it (there might be 0,1, or 2 values).

then we can eliminate the `hasDeletes` since this code would have no effect for an empty array.
```
const destructiveChangesTypes = cs.getTypesOfDestructiveChanges();
              destructiveChangesTypes.map((destructiveChangesType) => {
```

src/convert/metadataConverter.ts
```diff
@@ -103,13 +102,19 @@ export class MetadataConverter {
           if (!isSource) {
             const manifestPath = join(packagePath, MetadataConverter.PACKAGE_XML_FILE);
             tasks.push(promises.writeFile(manifestPath, manifestContents));
-
             // For deploying destructive changes
             if (cs.hasDeletes) {
```
mshanemc, line 106: see comments on CS about hasDeletes and getTypesOfDestructiveChanges

src/resolve/sourceComponent.ts
```diff
@@ -131,8 +133,23 @@ export class SourceComponent implements MetadataComponent {
...
+    if (
+      destructiveChangeType === DestructiveChangesType.PRE ||
+      destructiveChangeType === DestructiveChangesType.POST ||
+      destructiveChangeType === true ||
+      destructiveChangeType === undefined
+    ) {
+      this.markedForDelete = true;
+      this.destructiveChangesType =
+        (destructiveChangeType as DestructiveChangesType) || DestructiveChangesType.POST;
```
mshanemc, line 149: if `true` is passed in, you're making an invalid type assertion `true as DestructiveChangesType`.

What all would the `else` condition handle?  `false` only?  If so, maybe invert the order and return early?   

then you could have this.markedForDelete = true (because it's not false)
and then everything except PRE would mean POST, right?

test/resolve/sourceComponent.test.ts
```diff
@@ -60,7 +60,7 @@ describe('SourceComponent', () => {
   it('should return correct markedForDelete status', () => {
     expect(COMPONENT.isMarkedForDelete()).to.be.false;
     try {
```
mshanemc, line 62: this would be a good place to try out the other set values (destructive types, true)

mshanemc (changes requested):
most of the comments are just style stuff and concision or ideas.  

`setMarkedForDelete` and `getTypesOfDestructiveChanges ` are the only actual problems I saw.

mshanemc (changes requested):
most of the comments are just style stuff and concision or ideas.  

`setMarkedForDelete` and `getTypesOfDestructiveChanges ` are the only actual problems I saw.

src/convert/metadataConverter.ts
```diff
@@ -203,4 +202,13 @@ export class MetadataConverter {
...
   }
+
+  private convertTypeToManifest(manifestFileName: DestructiveChangesType): string {
+    if (manifestFileName === DestructiveChangesType.POST) {
+      return MetadataConverter.DESTRUCTIVE_CHANGES_POST_XML_FILE;
+    } else if (manifestFileName === DestructiveChangesType.PRE) {
+      return MetadataConverter.DESTRUCTIVE_CHANGES_PRE_XML_FILE;
+    }
+    return MetadataConverter.PACKAGE_XML_FILE;
+  }
```
shetzel, line 213: Bunch of nits:
1. I don't like the name of this method.  "Type" much more commonly refers to a metadata type in SDR.  This overloads the term which affects maintainability.  This is really about getting the destructive changes file name only right?
2. I don't think it's possible for `PACKAGE_XML_FILE` to ever be returned with how it's being called in the class.
3. The name of the arg should probably be something like, `destructiveChangesType`.

shetzel (changes requested):

src/collections/componentSet.ts
```diff
@@ -171,23 +175,57 @@ export class ComponentSet extends LazyCollection<MetadataComponent> {
...
+    if (options.destructivePost) {
+      const manifestPost = await manifestResolver.resolve(options.destructivePost);
+      for (const component of manifestPost.components) {
+        addComponent(
+          new SourceComponent({
+            type: component.type,
+            name: component.fullName,
+          }),
+          DestructiveChangesType.POST
+        );
```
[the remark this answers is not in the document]
shetzel:
```suggestion
const resolveDestructiveChanges = (option, type) => {
  const manifest = await manifestResolver.resolve(option);
  for (const comp of manifest.components) {
      addComponent(new SourceComponent({ type: comp.type, name: comp.fullName }), type);
  }
};
if (options.destructivePre) {
  resolveDestructiveChanges(options.destructivePre, DestructiveChangesType.PRE);
}
if (options.destructivePost) {
  resolveDestructiveChanges(options.destructivePost, DestructiveChangesType.POST);
}
```

shetzel (approved):

PR #452, opened 2021-09-16
fix: better support for non-nested inFolder types

shetzel:
### What does this PR do?
Reset the type after each manifest member entry.  Fixes the recent regression.  Note however that nested inFolder type metadata is still not supported properly but will be asap.

https://github.com/forcedotcom/cli/issues/1165
https://github.com/forcedotcom/cli/issues/1180
@W-9882522@

### Functionality Before
Deploy Error:
An  object  'emailTemplateDir/emailTemplateName' of  type EmailTemplate was named in package.xml, but was not found in zipped  directory

### Functionality After
Successful convert and deploy

### DIFF
--- a/src/resolve/manifestResolver.ts
+++ b/src/resolve/manifestResolver.ts
@@ -62,14 +62,15 @@ export class ManifestResolver {
 
     for (const typeMembers of packageTypeMembers) {
       const typeName = typeMembers.name;
-      let type = this.registry.getTypeByName(typeName);
+      const type = this.registry.getTypeByName(typeName);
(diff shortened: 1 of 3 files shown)


WillieRuemmele (approved):
QA'd with `-m EmailFolder` `-m EmailTemplate` `-m EmailFolder:<name>` `-m EmailTemplate:<name>` `-p <path/to/emails>` `-x <manifest including (EmailFolder | EmailTemplate)`

should also fix [nuts](https://github.com/salesforcecli/plugin-source/pull/208)

mshanemc (approved):

PR #453, opened 2021-09-16
fix(registry.json): registry update for ManagedTopics

shetzel:
### What does this PR do?
Updates the registry for ManagedTopics so that ManagedTopic children will not be included in the manifest during a source convert.

https://github.com/forcedotcom/cli/issues/1192
@W-9910605@

### Functionality Before
errors with: `An object 'topicName' of type ManagedTopic was named in package.xml, but was not found in zipped directory.`

### Functionality After
success

### DIFF
--- a/src/registry/registry.json
+++ b/src/registry/registry.json
@@ -1381,8 +1381,6 @@
           "managedtopic": {
             "id": "managedtopic",
             "name": "ManagedTopic",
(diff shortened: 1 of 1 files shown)


WillieRuemmele (approved):
QA'd with Steve, convert, deploy, retrieve all work with `ManagedTopic`

PR #455, opened 2021-09-22
fix: adds support for nested InFolder metadata types

shetzel:
### What does this PR do?
Properly resolves nested InFolder metadata types and their respective folder container types.
Report, ReportFolder, Document, DocumentFolder, Dashboard, DashboardFolder, EmailTemplate, EmailFolder

@W-9739746@
https://github.com/forcedotcom/cli/issues/1173
https://github.com/forcedotcom/cli/issues/1112

### Functionality Before
nested InFolder types failed to convert, deploy and retrieve

### Functionality After
successful convert, deploy, retrieve

Testing Notes:
Convert, deploy and retrieve the metadata types listed above as well as Territory2, Territory2Model, Territory2Rule, Territory2Type

### DIFF
--- a/src/registry/registry.json
+++ b/src/registry/registry.json
@@ -1431,6 +1431,7 @@
       "name": "Territory2Model",
       "suffix": "territory2Model",
       "directoryName": "territory2Models",
+      "inFolder": false,
       "folderType": "territory2model"
     },
     "territory2rule": {
@@ -1438,6 +1439,7 @@
       "name": "Territory2Rule",
       "suffix": "territory2Rule",
       "directoryName": "rules",
+      "inFolder": false,
       "folderType": "territory2model"
     },
     "territory2": {
@@ -1445,6 +1447,7 @@
       "name": "Territory2",
       "suffix": "territory2",
       "directoryName": "territories",
+      "inFolder": false,
       "folderType": "territory2model"
     },
     "campaigninfluencemodel": {
--- a/src/registry/registry.json
+++ b/src/registry/registry.json
@@ -1381,8 +1381,6 @@
           "managedtopic": {
             "id": "managedtopic",
             "name": "ManagedTopic",
-            "xmlElementName": "managedTopic",
-            "uniqueIdElement": "fullName",
             "directoryName": "managedTopics",
             "suffix": "managedTopic"
           }
--- a/src/registry/registryAccess.ts
+++ b/src/registry/registryAccess.ts
@@ -14,6 +14,9 @@ import { MetadataRegistry, MetadataType } from './types';
 export class RegistryAccess {
   private registry: MetadataRegistry;
 
+  private strictFolderTypes: MetadataType[];
+  private folderContentTypes: MetadataType[];
+
   constructor(registry: MetadataRegistry = defaultRegistry) {
     this.registry = registry;
   }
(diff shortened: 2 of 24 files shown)


src/registry/registryAccess.ts
```diff
@@ -71,9 +74,31 @@ export class RegistryAccess {
...
+   * Query for the types that have the folderContentType property defined.
+   * E.g., reportFolder, dashboardFolder, documentFolder, emailFolder
+   * @see {@link MetadataType.folderContentType}
+   *
+   * @returns An array of metadata type objects that have folder content
+   */
+  public getFolderContentTypes(): MetadataType[] {
+    if (!this.folderContentTypes) {
+      this.folderContentTypes = [];
+      for (const type of Object.values(this.registry.types)) {
```
WillieRuemmele, line 95: Do you think this is why the registry has so many other properties? To avoid the perf hit of iterating over every type just to find some that have a property defined? Would it make sense to add a `folderContentTypes` property to the registry and just access it directly?

shetzel: By "so many other properties" are you referring to the indexes like suffixes, strictDirectoryNames, and childTypes?  If so, yeah that's for better perf.  There are only 4 folderContentTypes so I don't think it's worth indexing them.  Building the array once makes sense to me though.  I think RegistryAccess could use more memoization logic like this.

src/registry/registry.json
```diff
@@ -1429,20 +1429,23 @@
...
       "directoryName": "rules",
+      "inFolder": false,
       "folderType": "territory2model"
     },
     "territory2": {
       "id": "territory2",
       "name": "Territory2",
       "suffix": "territory2",
       "directoryName": "territories",
+      "inFolder": false,
```
shetzel, line 1448: Adding these back since they are returned by a describemetadata call, and that's what we should be using as the source of truth for type definitions.

src/registry/registryAccess.ts
```diff
@@ -71,9 +74,31 @@ export class RegistryAccess {
    * @returns An array of metadata type objects that require strict parent folder names
    */
   public getStrictFolderTypes(): MetadataType[] {
-    return Object.values(this.registry.strictDirectoryNames).map(
-      (typeId) => this.registry.types[typeId]
-    );
+    if (!this.strictFolderTypes) {
+      this.strictFolderTypes = Object.values(this.registry.strictDirectoryNames).map(
+        (typeId) => this.registry.types[typeId]
+      );
```
shetzel, line 80: We only need to do this once.  The registry data can't change once RegistryAccess is created.

src/registry/registryAccess.ts
WillieRuemmele: yeah exactly. I'd prefer not to iterate over the whole registry to find four entries, multiple times... but we could do this optimization later

shetzel: It's only done once, not multiple times.

WillieRuemmele (approved):

src/registry/registry.json
mshanemc: source of untruth

src/resolve/manifestResolver.ts
```diff
@@ -64,16 +64,46 @@ export class ManifestResolver {
...
+    members: string[]
+  ): boolean {
+    // Quick short-circuit for non-folderTypes
+    if (!type.folderType) {
+      return false;
+    }
+
+    const isInFolderType = type.inFolder;
+    const isNestedInFolder =
+      !fullName.includes('/') || members.some((m) => m.includes(`${fullName}/`));
```
mshanemc, line 104: can windows users put `\` or `\\` in their manifest?

shetzel: I don't think that's valid.  It's not a file path, which would be OS specific.  It's the server-assigned API name.

PR #456, opened 2021-09-22
Sm/windows paths

mshanemc:
os-agnostic paths in tests and nestedFolderParsing


xyc (approved):

shetzel (changes requested):
I think this PR should be closed.  FullNames must use forward slashes and the other `join` unit test fixes were already made in the sh/nested-inFolder-types-support branch

PR #460, opened 2021-09-27
Sm/registry completeness

mshanemc:
### What does this PR do?
UT to download the mdapi coverage report and compare it to the SDR registry.
validation tests for registry
1. suffixes, strictDirs are present for all types
2. only one non-strictDir for types that share a suffix

CI to run only the completeness nightly (to quickly catch new types in the cov report that we didn't explicitly ignore)

Scripts for registry modifications from describe on real orgs, with support for ignored features/settings/types, based on the above completeness validation logic.

@W-9935205@
Screen Shot 2021-09-27 at 5 45 34 PM


test/registry/registryValidation.test.ts
```diff
@@ -0,0 +1,94 @@
...
+        ).body
+      ) as CoverageObject;
+
+      // these include child types
+      const metadataApiTypes = Object.entries(metadataCoverage.types).filter(
+        ([key, value]) => value.channels.metadataApi && !key.endsWith('Settings')
+      );
+      expect(metadataApiTypes.length).to.be.greaterThan(200);
+
+      describe('generated type specs', () => {
```
shetzel, line 57: What if we generated 2 arrays of type names and just did a deep compare?

test/registry/registryValidation.test.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  describe('every type from metadata coverage is in the SDR registry', () => {
+    before(async function () {
+      this.timeout(10000);
+      const metadataCoverage = JSON.parse(
+        (
+          await got(
+            `https://mdcoverage.secure.force.com/services/apexrest/report?version=${apiVersionMax}`
+          )
+        ).body
```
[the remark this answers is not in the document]
mshanemc: The const at the top would need to change.  It's just a question of when (sb0?  r2?)

You could call it without a version, but then this test becomes broken as soon as the coverageReport changes.  Maybe that's what we want?

I intended this to start some brainstorming about how it "should" work

test/registry/registryValidation.test.ts
mshanemc: I tried that but order matters on a deep compare.

randi274 (approved):

test/registry/registryValidation.test.ts
shetzel: `.to.have.deep.members()` shouldn't care about order...or just sort the arrays.  It's a nit.  Just don't like the dummy test.

mshanemc: oh, I never tried that before.  nice.

PR #461, opened 2021-09-28
fix: resolve CustomFieldTranslations properly

shetzel:
### What does this PR do?
If a CustomFieldTranslation was resolved before the parent CustomObjectTranslation it would not set the parent SourceComponent content, which caused problems during transformation.

@W-9946398@

### Functionality Before
Doesn't convert the field translations properly

### Functionality After
CustomFieldTranslations are converted properly

### DIFF
--- a/test/resolve/adapters/decomposedSourceAdapter.ts
+++ b/test/resolve/adapters/decomposedSourceAdapter.ts
@@ -11,15 +11,12 @@ import {
   decomposedtoplevel,
   mockRegistryData,
   xmlInFolder,
-  matchingContentFile,
 } from '../../mock/registry';
-import { assert, expect } from 'chai';
-import { VirtualTreeContainer, SourceComponent } from '../../../src';
+import { expect } from 'chai';
+import { VirtualTreeContainer, SourceComponent, MetadataType } from '../../../src';
 import { RegistryTestUtil } from '../registryTestUtil';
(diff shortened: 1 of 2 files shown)


test/resolve/adapters/decomposedSourceAdapter.ts
```diff
@@ -39,6 +36,27 @@ describe('DecomposedSourceAdapter', () => {
     );
   });
 
+  it('should return parent SourceComponent when given a child xml of non-addressable type', () => {
+    const nonAddressableType = JSON.parse(
+      JSON.stringify(mockRegistryData.types.decomposedtoplevel)
```
WillieRuemmele, line 41: why `parse` and `stringify` shouldn't those cancel each other out?

shetzel: to make a deep clone of the object so I can modify it and not affect other tests.

mshanemc (approved):

WillieRuemmele (approved):

WillieRuemmele:
using [this](https://github.com/iowillhoit/live-chat-transcript/tree/main/live-chat-project) repo I was able to `source:convert -d test` and then inspecting `test/LiveChatTranscript-de.objectTranslation` and all other files were written with the translations correctly

PR #465, opened 2021-10-02
fix: set correct status on file responses

shetzel:
### What does this PR do?
Sets the status on FileResponses to "Created" when they are not in the ComponentSet used to create the retrieve request.

@W-9921689@

### Functionality Before
All file responses returned by `RetrieveResult.getFileResponses()` would have a status of "Changed", even if they were new files for the local project.

### Functionality After
Retrieved files that are not on the local file system have a status of "Created".

### DIFF
--- a/src/client/metadataApiRetrieve.ts
+++ b/src/client/metadataApiRetrieve.ts
@@ -30,15 +30,29 @@ export type MetadataApiRetrieveOptions = MetadataTransferOptions &
 
 export class RetrieveResult implements MetadataTransferResult {
   public readonly response: MetadataApiRetrieveStatus;
+  // The ComponentSet of retrieved components.
   public readonly components: ComponentSet;
 
-  constructor(response: MetadataApiRetrieveStatus, components: ComponentSet) {
+  // The ComponentSet of just the components on the local file system.
+  private localComponents: ComponentSet;
+  private fileResponses: FileResponse[];
+
+  constructor(
+    response: MetadataApiRetrieveStatus,
+    components: ComponentSet,
+    localComponents?: ComponentSet
+  ) {
     this.response = response;
     this.components = components;
+    this.localComponents = new ComponentSet(localComponents?.getSourceComponents());
   }
 
   public getFileResponses(): FileResponse[] {
-    const responses: FileResponse[] = [];
+    if (this.response && this.fileResponses) {
+      return this.fileResponses;
+    }
+
+    this.fileResponses = [];
 
     // construct failures
     if (this.response.messages) {
@@ -49,7 +63,7 @@ export class RetrieveResult implements MetadataTransferResult {
         const matches = message.problem.match(/.+'(.+)'.+'(.+)'/);
         if (matches) {
           const [typeName, fullName] = matches.slice(1);
-          responses.push({
+          this.fileResponses.push({
             fullName,
             type: typeName,
             state: ComponentStatus.Failed,
@@ -57,7 +71,7 @@ export class RetrieveResult implements MetadataTransferResult {
             problemType: 'Error',
           });
         } else {
-          responses.push({
+          this.fileResponses.push({
             fullName: '',
             type: '',
             problemType: 'Error',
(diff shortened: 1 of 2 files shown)


src/client/metadataApiRetrieve.ts
```diff
@@ -175,7 +191,7 @@ export class MetadataApiRetrieve extends MetadataTransfer<
 
     await this.maybeSaveTempDirectory('source', components);
 
-    return new RetrieveResult(result, components);
+    return new RetrieveResult(result, components, this.components);
```
mshanemc, line 194: suggestion: `components` and `this.components` is confusing. Could `components` be `componentsFromZip`

src/client/metadataApiRetrieve.ts
```diff
@@ -30,15 +30,29 @@ export type MetadataApiRetrieveOptions = MetadataTransferOptions &
...
+    localComponents?: ComponentSet
+  ) {
     this.response = response;
     this.components = components;
+    this.localComponents = new ComponentSet(localComponents?.getSourceComponents());
   }
 
   public getFileResponses(): FileResponse[] {
-    const responses: FileResponse[] = [];
+    if (this.response && this.fileResponses) {
```
mshanemc, line 51: Why move this up to the instance?  Do some our consumers call `getFileResponses` multiple times and you're trying to save work?

src/client/metadataApiRetrieve.ts
mshanemc: especially with the RetrieveResult and the MetadataApiRetrieve classes being in the same file (because MetadataApiRetrieve has its own definition of `this.components`, which was [I think?] the componentSet that retrieve was called with)

src/client/metadataApiRetrieve.ts
```diff
@@ -30,15 +30,29 @@ export type MetadataApiRetrieveOptions = MetadataTransferOptions &
...
+  // The ComponentSet of retrieved components.
   public readonly components: ComponentSet;
 
-  constructor(response: MetadataApiRetrieveStatus, components: ComponentSet) {
+  // The ComponentSet of just the components on the local file system.
+  private localComponents: ComponentSet;
+  private fileResponses: FileResponse[];
+
+  constructor(
+    response: MetadataApiRetrieveStatus,
```
mshanemc, line 41: can it use the constructor shorthand form?
https://www.typescriptlang.org/docs/handbook/typescript-tooling-in-5-minutes.html#classes

mshanemc (approved):
approved with non-blocking comments.

WillieRuemmele (approved):
Agree with Shane's comments, but still LGTM

src/client/metadataApiRetrieve.ts
shetzel: Yes, it's called multiple times and there's no reason to do the work more than once after we get the response.

src/client/metadataApiRetrieve.ts
shetzel: `components` == the `ComponentSet` retrieved from the org.  Calling `getSourceComponents()` on this comp set would yield every component in the zip tree container.
`this.components` == the `ComponentSet` used to build the request to send to the org.  Calling `getSourceComponents()` on this comp set yields only the components on the local file system (technically, the TreeContainer).  The difference between the 2 comp sets are the "Created" files.

I can rename to be more clear.

shetzel: Renaming would be a breaking change so I instead updated the constructor docs to state what `components` refers to.

src/client/metadataApiRetrieve.ts
```diff
@@ -30,15 +30,29 @@ export type MetadataApiRetrieveOptions = MetadataTransferOptions &
...
+    components: ComponentSet,
+    localComponents?: ComponentSet
+  ) {
     this.response = response;
     this.components = components;
+    this.localComponents = new ComponentSet(localComponents?.getSourceComponents());
   }
 
   public getFileResponses(): FileResponse[] {
```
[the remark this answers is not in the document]
shetzel: I'm not opposed to this, but:

1. there is some symmetry between metadataApiRetrieve and metadataApiDeploy so should this apply to both?
2. that would build 2 arrays and a concat into another, which is less performant than just building 1 array.  Probably not noticeable but is it worth it?

src/client/metadataApiRetrieve.ts
```diff
@@ -29,21 +29,22 @@ export type MetadataApiRetrieveOptions = MetadataTransferOptions &
...
 
+  /**
+   * @param response The metadata retrieve response from the server
+   * @param components The ComponentSet of retrieved source components
+   * @param localComponents The ComponentSet used to create the retrieve request
+   */
   constructor(
-    response: MetadataApiRetrieveStatus,
-    components: ComponentSet,
+    readonly response: MetadataApiRetrieveStatus,
```
mshanemc, line 44:
```suggestion
    public readonly response: MetadataApiRetrieveStatus,
```
default public, but probably good to make it explicitly clear that, for example, you can't rename it without breaking stuff.

src/client/metadataApiRetrieve.ts
```diff
@@ -29,21 +29,22 @@ export type MetadataApiRetrieveOptions = MetadataTransferOptions &
...
+    readonly components: ComponentSet,
```
mshanemc, line 45:
```suggestion
    public readonly components: ComponentSet,
```

PR #466, opened 2021-10-04
fix mismatches and missing properties in registry

jayree:

#<Insert GitHub Issue>, @<Insert GUS WI>@

### Functionality Before

mismatches and missing properties in registry

### Functionality After

updated registry


jayree:
@iowillhoit this is an example:
```<?xml version="1.0" encoding="UTF-8"?>
<Workflow xmlns="http://soap.sforce.com/2006/04/metadata">
    <send>
        <fullName>sendaction</fullName>
        <action>Send</action>
        <label>sendaction</label>
        <language>en_US</language>
        <protected>false</protected>
    </send>
</Workflow>

PR #474, opened 2021-10-11
fix: keys split into 2 strings, which allows # in fullName

mshanemc:
### What does this PR do?
allow `#` in fullNames for components

#<Insert GitHub Issue>, @<Insert GUS WI>@
@W-5185915@

### Functionality Before

For a component with `#` in its name (see example on the WI), SDR would call string.split('#') and get confused.  When trying to pull, it wouldn't be able to find matching metadata

### Functionality After
finds metadata (has to split into two strings on the first instance of the delimiter)


WillieRuemmele (approved):
let's hope there's never a metadata type with `#` in it

mshanemc:
> let's hope there's never a metadata type with `#` in it
<shudders> 

`Custom#Labels`

PR #482, opened 2021-10-20
fix: migrate registry.json from fork

WillieRuemmele:
moving into branch off of main

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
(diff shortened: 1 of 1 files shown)


src/registry/registry.json
```diff
@@ -2884,7 +2915,8 @@
     "accessControlPolicies": "accesscontrolpolicy",
     "integrationHub": "integrationhubsettings",
     "ObjectHierarchyRelationship": "objecthierarchyrelationship",
-    "IndustriesManufacturingSettings": "industriesmanufacturingsettings"
+    "IndustriesManufacturingSettings": "industriesmanufacturingsettings",
```
maggiben, line 2918: This one: `IndustriesManufacturingSettings` is bad, bad boy, I caused some trouble in the past because of it's extension name.

PR #495, opened 2021-11-03
Sm/potential-filenames

mshanemc:
### What does this PR do?
give it a name/type, and you get back an array of filePaths.

This won't give all the filenames for decomposed types (that would require retrieving the actual parent xml) but should provide enough of the filePath (for example, where a field would go) to figure out if the forceignore would ignore it.

@W-10095309@

Testing:
it's a library, so probably the easiest way to test this is to write some more UT.  See 
/Users/shane.mclaughlin/eng/source-deploy-retrieve/test/utils/filePathGenerator.test.ts

for examples, but try some other things that I didn't (and feel free to contribute them in a PR)


src/utils/filePathGenerator.ts
```diff
@@ -0,0 +1,114 @@
...
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import { join, sep } from 'path';
+import { MetadataComponent } from '../resolve/types';
+import { META_XML_SUFFIX } from '../common/constants';
+import { RegistryAccess } from '../registry/registryAccess';
+import { registry } from '..';
+const registryAccess = new RegistryAccess();
+
+export const filePathsFromMetadataComponent = (
```
shetzel, line 14: This is public, so needs js doc, particularly calling out the note you made in the PR description.

shetzel (approved):

PR #516, opened 2021-12-20
Sm/sdr-mdapi-zip

mshanemc:
### What does this PR do?
deploy via mdapiDir or a zip file

test: I am baffled by source-deploy-retrieve/test/mock/client/transferOperations.ts and don't know how to get my additional deploy parameters onto it.

@W-9603711@

### DIFF
--- a/src/client/metadataApiDeploy.ts
+++ b/src/client/metadataApiDeploy.ts
@@ -5,7 +5,6 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 import { basename, dirname, extname, join } from 'path';
-import { Stream } from 'stream';
 import { isString } from '@salesforce/ts-types';
 import { create as createArchive } from 'archiver';
 import * as fs from 'graceful-fs';
@@ -15,6 +14,7 @@ import { normalizeToArray } from '../utils';
 import { ComponentSet } from '../collections';
 import { registry } from '../registry';
 import { MissingJobIdError } from '../errors';
+import { stream2buffer } from '../convert/streams';
 import { MetadataTransfer, MetadataTransferOptions } from './metadataTransfer';
 import {
   AsyncResult,
@@ -26,7 +26,6 @@ import {
   MetadataTransferResult,
 } from './types';
 import { DiagnosticUtil } from './diagnosticUtil';
-
 export class DeployResult implements MetadataTransferResult {
   public readonly response: MetadataApiDeployStatus;
   public readonly components: ComponentSet;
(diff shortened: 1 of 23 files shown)


src/client/metadataApiDeploy.ts
```diff
@@ -313,4 +322,31 @@ export class MetadataApiDeploy extends MetadataTransfer<MetadataApiDeployStatus,
   protected async post(result: MetadataApiDeployStatus): Promise<DeployResult> {
     return new DeployResult(result, this.components);
   }
+
+  private async getZipBuffer(): Promise<Buffer> | undefined {
```
shetzel, line 326: When does this return undefined?  If it does return undefined it would still need to be `Promise<void>` right?  Seems like either `Promise<Buffer>` or `Promise<Optional<Buffer>>` is the return type.

src/client/metadataApiDeploy.ts
```diff
@@ -313,4 +322,31 @@ export class MetadataApiDeploy extends MetadataTransfer<MetadataApiDeployStatus,
...
+    if (this.options.mdapiPath) {
+      if (!fs.existsSync(this.options.mdapiPath) || !fs.lstatSync(this.options.mdapiPath).isDirectory()) {
+        throw new Error(`Deploy directory ${this.options.mdapiPath} does not exist or is not a directory`);
+      }
```
shetzel, line 330: Should we also ensure a manifest is in the dir?  it would save an http call if it wasn't.  Same for the zipPath option.

test/client/metadataApiDeploy.test.ts
```diff
@@ -35,6 +35,9 @@ describe('MetadataApiDeploy', () => {
 
   describe('Lifecycle', () => {
     describe('start', () => {
+      it('should not convert zip, but read from fs', async () => {});
+      it('should not mdapiDir, but generate zip buffer from it', async () => {});
```
shetzel, line 39: You'll have to modify `stubMetadataDeploy()` in `transferOperations.ts` to stub all the additional fs, zip, and streaming code you added in `getZipBuffer` and return the stubs from `stubMetadataDeploy`.  You'll also have to modify that function to return an `operation` with an instance of `MetadataApiDeploy` created with the mdapiPath and zipPath options.  Your test would then call `operation.start()` like the others, and you would verify stubs were called or not called with the expected args.

src/client/metadataApiDeploy.ts
mshanemc: I'm just gonna throw at the bottom

shetzel (approved):

PR #517, opened 2021-12-22
fix: can retrive COFT with COT automatically

WillieRuemmele:
### What does this PR do?
fixes retrieval of COFT

https://github.com/forcedotcom/cli/issues/1233, https://github.com/forcedotcom/cli/issues/1241, https://github.com/forcedotcom/cli/issues/1262, @W-10222972@

### Functionality Before
`retrieve -m CustomFieldTranslation` without a CustomObjectTranslation in the org would only pull the CustomFieldTranslation files, which were invalid by themselves without a CustomObjectTranslation parent file

### Functionality After
a retrieve will retrieve both COFT/COT and write an empty COT if one doesn't exist in the org

### DIFF
--- a/src/convert/transformers/decomposedMetadataTransformer.ts
+++ b/src/convert/transformers/decomposedMetadataTransformer.ts
@@ -86,7 +86,7 @@ export class DecomposedMetadataTransformer extends BaseMetadataTransformer {
              CustomObjectFieldTranslations are only addressable through their parent, and require a
              CustomObjectTranslation file to be present
              */
-            if (childType.requiresParent && composedMetadata.length <= 2) {
+            if (childType.unaddressableWithoutParent && composedMetadata.length <= 2) {
               parentXmlObject = {
                 [component.type.name]: '',
               };
--- a/src/convert/transformers/decomposedMetadataTransformer.ts
+++ b/src/convert/transformers/decomposedMetadataTransformer.ts
@@ -13,6 +13,7 @@ import { META_XML_SUFFIX, SourcePath, XML_NS_KEY, XML_NS_URL } from '../../commo
 import { ComponentSet } from '../../collections';
 import { DecompositionState } from '../convertContext';
 import { DecompositionStrategy } from '../../registry';
+import { normalizeToArray } from '../../utils';
 import { BaseMetadataTransformer } from './baseMetadataTransformer';
 
 export class DecomposedMetadataTransformer extends BaseMetadataTransformer {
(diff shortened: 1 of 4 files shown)


src/convert/transformers/decomposedMetadataTransformer.ts
```diff
@@ -74,6 +75,30 @@ export class DecomposedMetadataTransformer extends BaseMetadataTransformer {
...
+             composedMetadata is a representation of the parent's xml
+             if there is no CustomObjectTranslation in the org, the composedMetadata will be 2 entries
+             the xml declaration, and a fields attribute, which points to the child CustomObjectFieldTranslation
+             because CustomObjectFieldTranslation is the only metadata type with 'requiresParent' = true we can
+             calculate if a CustomObjectTranslation was retrieved from the org (composedMetadata.length > 2), or,
+             if we'll have to write an empty CustomObjectTranslation file (composedMetadata.length <=2).
+             CustomObjectFieldTranslations are only addressable through their parent, and require a
+             CustomObjectTranslation file to be present
+             */
+            if (childType.requiresParent && composedMetadata.length <= 2) {
```
WillieRuemmele, line 89: this could use the `isAddressable` field, because COFT is the only metadata type to set either of those. It just seemed like the naming and intention could become confusing if I used `isAddressable` 🤷‍♂️

src/convert/transformers/decomposedMetadataTransformer.ts
```diff
@@ -61,7 +62,7 @@ export class DecomposedMetadataTransformer extends BaseMetadataTransformer {
       const childTypeId = type.children?.directories[tagKey];
       if (childTypeId) {
         const childType = type.children.types[childTypeId];
-        const tagValues = Array.isArray(tagValue) ? tagValue : [tagValue];
+        const tagValues = normalizeToArray(tagValue);
```
mshanemc, line 65: +1

src/registry/types.ts
```diff
@@ -25,6 +25,7 @@ export interface MetadataRegistry {
  * Metadata type definition in the registry.
  */
 export interface MetadataType {
+  requiresParent?: boolean;
```
mshanemc, line 28: add the jsdoc for that the option does, like all the others (if we decide to use the new option).  Otherwise, modify isAddressable's explanation

src/convert/transformers/decomposedMetadataTransformer.ts
mshanemc: I'm gonna defer to Steve's opinion on this one.

My question is whether it's valid to have `isAddressable:false` without `requiresParent:true`, or vice versa?  

Or would you say that what we really meant was `unaddressableWithoutParent` ?

mshanemc (approved):
I'd ship it as is if it fixed the problem.  I'd like to see @shetzel opine on whether this should be a new property on `MetadataType` or not.

If it should be a new prop AND it's only valid in combination with isAddressable, then I'd want a validation to be added for that in registryValidation (verify that no childTypes exist with an invalid combination of those properties AND that they can only be used on childTypes).

src/convert/transformers/decomposedMetadataTransformer.ts
shetzel: I know there are more entities that are unaddressable, and that field is useful and makes sense so that one should remain.  I agree with Shane that `unaddressableWithoutParent` is better.  COT and COFT are unfortunate anomalies that hopefully will never happen again.  `unaddressableWithoutParent` is the nicest description of this disastrous pair.

shetzel (approved):

PR #518, opened 2021-12-28
fix: wrap error from fast-xml-parser when invalid xml parsed

WillieRuemmele:
### What does this PR do?
when the `fast-xml-parser` library throws an error due to invalid xml, we will parse that error, and add the file that caused the error. 

given a custom labels xml file
```
<?xml version="1.0" encoding="UTF-8"?>
</CustomLabels xmlns="http://soap.sforce.com/2006/04/metadata">
  <labels>
    <fullName>force_app_Label_1</fullName>
    <language>en_US</language>
    <protected>true</protected>
    <shortDescription>force-app Label 1</shortDescription>
    <value>force-app</value>
  </labels>
</CustomLabels>
```
 

note the opening CustomLabels tag is actually a closing tag

#[1261](https://github.com/forcedotcom/cli/issues/1261), @W-10119453@

### Functionality Before
`TypeError: Cannot read property 'addChild' of undefined`

### Functionality After
```
 Component conversion failed: error parsing dreamhouse-lwc/force-app/main/default/labels/CustomLabels.labels-meta.xml due to:
  message: Closing tag 'CustomLabels' can't have attributes or invalid starting..
  line: 2.
  code: InvalidTag.
```

### DIFF
--- a/src/resolve/sourceComponent.ts
+++ b/src/resolve/sourceComponent.ts
@@ -117,21 +117,7 @@ export class SourceComponent implements MetadataComponent {
     const xml = xmlFilePath ?? this.xml;
     if (xml) {
       const contents = (await this.tree.readFile(xml)).toString();
-      try {
-        return this.parse<T>(contents);
-      } catch (e) {
-        // only attempt validating once there's an error to avoid the performance hit of validating every file
-        const validation = validate(contents);
-        if (validation !== true) {
-          throw new LibraryError('invalid_xml_parsing', [
-            xml,
-            validation.err.msg,
-            validation.err.line.toString(),
-            validation.err.code,
-          ]);
-        }
-        throw e;
-      }
+      return this.parseAndValidateXML(contents, xml);
     }
     return {} as T;
   }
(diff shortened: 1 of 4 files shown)


src/resolve/sourceComponent.ts
```diff
@@ -116,17 +116,45 @@ export class SourceComponent implements MetadataComponent {
   public async parseXml<T = JsonMap>(xmlFilePath?: string): Promise<T> {
     const xml = xmlFilePath ?? this.xml;
     if (xml) {
-      const contents = await this.tree.readFile(xml);
-      return this.parse<T>(contents.toString());
+      const contents = (await this.tree.readFile(xml)).toString();
+      try {
+        return this.parse<T>(contents);
+      } catch (e) {
+        // only attempt validating once there's an error to avoid the performance hit of validating every file
```
mshanemc, line 123: +1

src/resolve/sourceComponent.ts
```diff
@@ -116,17 +116,45 @@ export class SourceComponent implements MetadataComponent {
...
     const xml = xmlFilePath ?? this.xml;
     if (xml) {
-      const contents = this.tree.readFileSync(xml);
-      return this.parse<T>(contents.toString());
+      const contents = this.tree.readFileSync(xml).toString();
+      try {
+        return this.parse<T>(contents);
+      } catch (e) {
+        // only attempt validating once there's an error to avoid the performance hit of validating every file
+        const validation = validate(contents);
```
mshanemc, line 147: I think having them both use a shared`private parseAndValidateXML()` would be more DRY.

mshanemc (changes requested):

src/resolve/sourceComponent.ts
WillieRuemmele: I had that for a while - but it was causing testing issues because `parse` was stubbed, but I got it working 👍

mshanemc (approved):

mshanemc:
QA notes

this was pretty hard to QA (SDR doesn't care about most xml or parse it on a source:convert or deploy, including apexClasses or object/field stuff).

PR #520, opened 2022-01-03
fix: add variable polling times based on comp. set

WillieRuemmele:
What does this PR do?
adds variable polling frequency to the `pollStatus` method when no value is passed for the frequency

I ended up creating a piece-wise function to capture "small" vs "medium" vs "large" vs "larger" deploys - I'm very open to changing the boundaries and frequencies each should return. I did notice a difference between 250ms and 100ms

@W-9586672@

### DIFF
--- a/src/client/metadataTransfer.ts
+++ b/src/client/metadataTransfer.ts
@@ -95,8 +95,8 @@ export abstract class MetadataTransfer<Status extends MetadataRequestStatus, Res
     }
     // from the overloaded methods, there's a possibility frequency/timeout isn't set
     // guarantee frequency and timeout are set
-    pollingOptions.frequency = pollingOptions.frequency ?? Duration.milliseconds(this.calculatePollingFrequency());
-    pollingOptions.timeout = pollingOptions.timeout ?? Duration.minutes(60);
+    pollingOptions.frequency ??= Duration.milliseconds(this.calculatePollingFrequency());
+    pollingOptions.timeout ??= Duration.minutes(60);
 
     const pollingClient = await PollingClient.create(pollingOptions);
 
--- a/src/client/metadataTransfer.ts
+++ b/src/client/metadataTransfer.ts
@@ -14,6 +14,7 @@ import { MetadataConverter, SfdxFileFormat } from '../convert';
 import { MetadataTransferError } from '../errors';
 import { ComponentSet } from '../collections';
 import { AsyncResult, MetadataRequestStatus, MetadataTransferResult, RequestStatus } from './types';
+
 export interface MetadataTransferOptions {
   usernameOrConnection: string | Connection;
   components?: ComponentSet;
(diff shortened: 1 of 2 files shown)


src/client/metadataTransfer.ts
```diff
@@ -92,6 +93,10 @@ export abstract class MetadataTransfer<Status extends MetadataRequestStatus, Res
     if (isNumber(timeout)) {
       pollingOptions.timeout = Duration.seconds(timeout);
     }
+    // from the overloaded methods, there's a possibility frequency/timeout isn't set
+    // guarantee frequency and timeout are set
+    pollingOptions.frequency = pollingOptions.frequency ?? Duration.milliseconds(this.calculatePollingFrequency());
```
mshanemc, line 98:
```suggestion
    pollingOptions.frequency ??= Duration.milliseconds(this.calculatePollingFrequency());
```

src/client/metadataTransfer.ts
```diff
@@ -92,6 +93,10 @@ export abstract class MetadataTransfer<Status extends MetadataRequestStatus, Res
...
+    pollingOptions.timeout = pollingOptions.timeout ?? Duration.minutes(60);
```
mshanemc, line 99:
```suggestion
    pollingOptions.timeout ??= Duration.minutes(60);
```

mshanemc (approved):

src/client/metadataTransfer.ts
```diff
@@ -216,6 +221,28 @@ export abstract class MetadataTransfer<Status extends MetadataRequestStatus, Res
...
   }
 
+  /**
+   * Based on the source components in the component set, it will return a polling frequency in milliseconds
+   */
+  private calculatePollingFrequency(): number {
+    const size = this.components?.getSourceComponents().toArray().length || 0;
+    // take a piece-wise approach to encapsulate discrete deployment sizes in polling frequencies that "feel" good when deployed
+    if (size === 0) {
+      // no component set size is possible for retrieve
```
randi274, line 231: What's the thought behind having a large polling frequency here? I'm not quite clear on the comment.

test/client/metadataTransfer.test.ts
```diff
@@ -163,7 +163,68 @@ describe('MetadataTransfer', () => {
...
+      );
+    });
+
+    it('should calculate polling frequency based on source components, 10 -> 100', async () => {
+      // @ts-ignore protected member access
+      env.stub(operation.components, 'getSourceComponents').returns({
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore only override length attribute
+        toArray: () => {
+          return { length: 1 };
```
randi274, line 187: Nit: should this return 10, based on the test name?

randi274 (commented):
Smart change to make! Just a few questions for my own understanding.

src/client/metadataTransfer.ts
WillieRuemmele: there is no way to calculate the size of SourceComponents in a retrieve request, in those cases, the `size` will be 0, and so to avoid polling excessively, we take a cautious approach and default to 1000ms

randi274: got it, thanks!

mshanemc:
QA Notes:
I'm QAing this via the plugin-source sibling PR.

using a small repo I've been using for mdapi:deploy
✅ `source:deploy` works with default wait
✅ `source:deploy` works with wait=0

using ebikes (n=99)
✅ `source:beta:push` works with default wait (still saw the off-by-one error pending W-10253654)
✅ `source:retrieve` works with default wait (I can see 500ms interval via the debug output)

✅ I can get the default (1000ms) on a retrieve by `force:source:retrieve -m ExternalDataSource` when there are none locally)

PR #553, opened 2022-01-18
Sm/return-ignores

mshanemc:
### What does this PR do?
make a more complete `forceIgnoredPaths` 

@W-10449080@

### Functionality Before

forceignored from metadataResolver or componentSet would only list resolved components that were ignored.  Commonly ignored files like `lwc/jsconfig.json` were not listed even though they were ignored because they weren't being recognized as components.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,7 +30,7 @@ jobs:
       node_version:
         description: version of node to run tests against
         type: string
-        default: 'latest'
+        default: 'lts'
       os:
         description: operating system to run tests on
         type: enum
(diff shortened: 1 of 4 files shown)


WillieRuemmele (approved):
I was a little worried this could have unintended consequences at first, but it's only filling out the `forceIgnoredPaths` list which should be safe enough 👍

shetzel (approved):

mshanemc:
abandoning because SDR doesn't crawl inside bundles for things like `__tests__` and we don't want it to for perf reasons.

PR #554, opened 2022-01-18
chore: migrate to schedule pipelines

cristiand391:
### What does this PR do?
Migrates to scheduled pipelines, adds 6 pipeline parameters for  scheduled triggers. Also adds `test-ts-update` workflow.

@W-10425244@

### Functionality Before

Scheduled workflows with schedules defined in config file

### Functionality After

Workflows only run when a parameter is provided (by an scheduled trigger defined in `Project settings > Triggers`, or circle API)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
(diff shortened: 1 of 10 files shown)


mshanemc (approved):
triggers are wrong in circle, but this PR is fine

mshanemc (approved):
yeah, I like this better.  

See slack thread for comment about just-nuts on plugins

randi274 (approved):

PR #557, opened 2022-01-25
fix: warn and convert on windows seps in .forceignore

WillieRuemmele:
### What does this PR do?
when the `.forceignore` finds a `\` it will warn the user that that format will no longer be supported, and will change it so that it'll work for them

@W-10464641@

### Functionality Before
with `force-app\main\default\classes\post.*` in the `.forceignore`, `sfdx force:source:deploy -p force-app/main/default/classes`

```
DEPLOY PROGRESS | ████████████████████████████████████████ | 8/8 Components

=== Deployed Source
FULL NAME                 TYPE       PROJECT PATH
────────────────────────  ─────────  ────────────────────────────────────────────────────────────────────
GeocodingService          ApexClass  force-app/main/default/classes/GeocodingService.cls
...
post                      ApexClass  force-app/main/default/classes/post.cls
post                      ApexClass  force-app/main/default/classes/post.cls-meta.xml
```

### Functionality After

<insert gif and/or summary>
```
(node:55500) Warning: Your .forceignore file incorrectly uses the backslash ("\") as a folder separator; it should use the slash ("/") instead. We currently accept both separators, but we plan to stop supporting the backslash soon.
DEPLOY PROGRESS | ████████████████████████████████████████ | 7/7 Components

=== Deployed Source
FULL NAME                 TYPE       PROJECT PATH
────────────────────────  ─────────  ────────────────────────────────────────────────────────────────────
GeocodingService          ApexClass  force-app/main/default/classes/GeocodingService.cls
GeocodingService          ApexClass  force-app/main/default/classes/GeocodingService.cls-meta.xml
...
TestSampleDataController  ApexClass  force-app/main/default/classes/TestSampleDataController.cls-meta.xml
```

### DIFF
--- a/src/resolve/forceIgnore.ts
+++ b/src/resolve/forceIgnore.ts
@@ -8,6 +8,7 @@
 import { dirname, join, relative } from 'path';
 import ignore, { Ignore } from 'ignore/index';
 import { readFileSync } from 'graceful-fs';
+import { Lifecycle } from '@salesforce/core';
 import { SourcePath } from '../common';
 import { searchUp } from '../utils/fileSystemHandler';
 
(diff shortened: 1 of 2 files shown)


src/resolve/forceIgnore.ts
```diff
@@ -16,16 +17,26 @@ export class ForceIgnore {
...
+      let contents = readFileSync(forceIgnorePath, 'utf-8');
       // check if file `.forceignore` exists
-      if (this.contents !== undefined) {
+      if (contents !== undefined) {
+        // check for windows style separators (\) and warn
+        if (contents.includes('\\')) {
+          const lifecycle = Lifecycle.getInstance();
+          // cannot await a method in a constructor
+          void lifecycle.emitWarning(
+            'Your .forceignore file incorrectly uses the backslash ("\\") as a folder separator; it should use the slash ("/") instead. We currently accept both separators, but we plan to stop supporting the backslash soon.'
```
WillieRuemmele, line 32: @jshackell-sfdc created and approved this message

src/resolve/forceIgnore.ts
```diff
@@ -16,16 +17,26 @@ export class ForceIgnore {
...
-  private readonly contents?: string;
   private DEFAULT_IGNORE: string[] = ['**/*.dup', '**/.*', '**/package2-descriptor.json', '**/package2-manifest.json'];
 
   public constructor(forceIgnorePath = '') {
     try {
-      this.contents = readFileSync(forceIgnorePath, 'utf-8');
+      let contents = readFileSync(forceIgnorePath, 'utf-8');
       // check if file `.forceignore` exists
-      if (this.contents !== undefined) {
+      if (contents !== undefined) {
```
jeffb-sfdc, line 26: Would `if (!contents)` be better here, or is this condition specifically for undefined?

mshanemc (approved):

src/resolve/forceIgnore.ts
mshanemc: @jeffb-sfdc empty string is falsy--but if someone really makes an empty forceignore file we'll want honor that above our defaults

(plus that would be a behavior change and someone'll get bit by it 🤪 )

maggiben:
QA:

:white_check_mark: Warns the user of incorrect backslash use:
image

PR #571, opened 2022-02-08
Component set from connection

shetzel:
What does this PR do?
This adds the capability to create a componentSet of metadataComponents from an org connection. This can/will be used to extend the source:manifest:create command to create a manifest file based on an org configuration.

@W-0@

Functionality Before
<insert gif and/or summary>

Functionality After
<insert gif and/or summary>


shetzel:
This is Jochen's work just in a PR I created.  I tested it by writing a quick command parameter implementation for the `force:source:manifest:create` command that uses this code to create a manifest based on an org.  I used the generated manifest for deploys and retrieves and it works great.

mshanemc (approved):

PR #572, opened 2022-02-09
rohitachar/ add new UiViewDefinition metadata type to registry

rshreekant:
### What does this PR do?

Add a new UiViewDefinition type to the Metadata Registry

@W-10495831@

### Functionality Before

sfdx force:source commands did not work for UiViewDefinition type

### Functionality After

sfdx force:source commands work for UiViewDefinition type

### DIFF
--- a/METADATA_SUPPORT.md
+++ b/METADATA_SUPPORT.md
@@ -545,3 +545,4 @@ v55 introduces the following new types.  Here's their current level of support
 - MktDataTranField
 - ConversationVendorFieldDef
 - InternalOrganization
+- UiViewDefinition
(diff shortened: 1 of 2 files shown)


mshanemc (approved):

TiagoNeve:
My package generation broke because of this pull request, please, is there a way to remove this UiViewDefinition metadata?

I ran the following command:
```
sf project generate manifest --from-org [org-name-hidden] -n teste.xml --output-dir manifest
```

And I got the following error.
```
Entity type 'UiViewDefinition' is not available in this organization
```

PR #573, opened 2022-02-10
feat: cache fileResponses on deployResult

mshanemc:
### What does this PR do?
cache getFileResponses inside DeployResult

why?
- plugin-source has to run this in multiple places (source tracking, formatters) and it could be slow for really big deploys
- calling it creates FS operations.  We'd like to only do once
- sometimes we modify the local FS after the retrieve (ex: pull with remote deletes OR source:delete). SDR throws when it can't find files that were in the fileReponses originally

secondary for @W-10666968@


shetzel (approved):

WillieRuemmele:
QA:
linked into `plugin-source`
debugged `beta:push` command with a breakpoint at the caching.
✅ : only calculated when the cache was empty
✅ : reused cache when it had already been calculated - used cache 3 times for 1 `push`

PR #575, opened 2022-02-14
Wr/memory usage

WillieRuemmele:
### What does this PR do?
Begins writing StaticResource files as soon as possible, rather than deferring them to the `WriteInfos` and finalizers

[1348](https://github.com/forcedotcom/cli/issues/1348), @W-10417338@

### Functionality Before
A large spike in memory usage as SDR writes all SR files, up to 3GB but really unrestricted

### Functionality After
no spike in memory usage when writing SR files

QA helpful notes:
clone the repo mentioned in the issue
open activity monitory
deploy everything in the project to a new SO
`retrieve -m StaticResource` while looking at activity monitory
notice huge spike in the memory right before it prints the table, the spinner even slows down

### DIFF
--- a/METADATA_SUPPORT.md
+++ b/METADATA_SUPPORT.md
@@ -488,9 +488,11 @@ v55 introduces the following new types.  Here's their current level of support
 |ExternalDataSrcDescriptor|❌|Not supported, but support could be added|
 |ExternalDataTranField|❌|Not supported, but support could be added|
 |ExternalDataTranObject|❌|Not supported, but support could be added|
+|FavoriteTransferDestination|❌|Not supported, but support could be added|
 |IndustriesAutomotiveSettings|✅||
 |IndustriesMfgServiceSettings|✅||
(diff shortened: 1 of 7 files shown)


mshanemc:
QA notes: 

✅ memory on this one static resource doesn't spike
✅ using node perf_hooks results in ~5% faster throughput compared to original

🏁 I made some tweaks to explore parallelization options (sm/tweaks-fp)  
✅ dropped ~25% (1s) from where this PR started (which was already faster than current)

let's break stuff...
😱 copied the static resource 10x and retrieved 9 of them simultaneously
✅ perf scales less than linearly (27.3s vs. 41.4 [9 * baseline from latest branch]).  There's some overlap happening as some wind down and the next ramp up, maybe? 
✅ even doing 9 of these still kept peak memory usage under 3GB (you'd see it hit that ceiling, drop, then come back up again)

mshanemc (approved):

PR #576, opened 2022-02-14
fix: add ComponentSetBuilder from plugin-source

WillieRuemmele:
### What does this PR do?
adds the `ComponentSetBuilder` the CLI team has been using in `plugin-source` to create CS

other consumers of SDR might find it useful as well as `sf` source plugin

@W-10657718@

### Functionality Before
N/A

### Functionality After
Use the component set builder to create component sets from a variety of sources, set api versions on the connection


src/collections/componentSetBuilder.ts
```diff
@@ -0,0 +1,152 @@
...
+    const { sourcepath, manifest, metadata, packagenames, apiversion, sourceapiversion } = options;
+    try {
+      if (sourcepath) {
+        logger.debug(`Building ComponentSet from sourcepath: ${sourcepath.join(', ')}`);
+        const fsPaths: string[] = [];
+        sourcepath.forEach((filepath) => {
+          if (!fs.existsSync(filepath)) {
+            throw new SfdxError(`The sourcepath "${filepath}" is not a valid source file path.`);
+          }
+          fsPaths.push(path.resolve(filepath));
```
mshanemc, line 57:
```suggestion
        const fsPaths =  sourcepath.map((filepath) => {
          if (!fs.existsSync(filepath)) {
            throw new SfdxError(`The sourcepath "${filepath}" is not a valid source file path.`);
          }
          return path.resolve(filepath);
```

src/collections/componentSetBuilder.ts
```diff
@@ -0,0 +1,152 @@
...
+      }
+
+      // Resolve manifest with source in package directories.
+      if (manifest) {
+        logger.debug(`Building ComponentSet from manifest: ${manifest.manifestPath}`);
+        const directoryPaths = options.manifest.directoryPaths;
+        logger.debug(`Searching in packageDir: ${directoryPaths.join(', ')} for matching metadata`);
+        componentSet = await ComponentSet.fromManifest({
+          manifestPath: manifest.manifestPath,
+          resolveSourcePaths: options.manifest.directoryPaths,
```
mshanemc, line 75: why not use `directoryPaths` since there's already a const?

mshanemc (changes requested):
2 minor changes, please

mshanemc (commented):
I don't have any problems with what this does.

My only concern is that we have @jayree 's PR open here https://github.com/salesforcecli/plugin-source/pull/423 and we want that in as well.  So either add that stuff if @shetzel  says it's "done" OR hold off on this and then copy-paste that when it is done

WillieRuemmele:
In draft until https://github.com/salesforcecli/plugin-source/pull/423 has been finalized

src/collections/componentSetBuilder.ts
```diff
@@ -0,0 +1,171 @@
...
+        logger.debug(`Searching for matching metadata in directories: ${directoryPaths.join(', ')}`);
+        const resolvedComponents = ComponentSet.fromSource({ fsPaths: directoryPaths, include: compSetFilter });
+        componentSet.forceIgnoredPaths = resolvedComponents.forceIgnoredPaths;
+        for (const comp of resolvedComponents) {
+          componentSet.add(comp);
+        }
+      }
+
+      // Resolve metadata entries with an org connection
+      if (org) {
```
shetzel, line 119: This is where I think we'll have to add some validation of `ComponentSetOptions`.  I can't think of a reason why someone would want to combine the `org` option with others, but line 121 below will clobber `componentSet` anyway.  Either the reassignment should change to be additive or we need to validate how people are combining `ComponentSetOptions`.

src/collections/componentSetBuilder.ts
```diff
@@ -0,0 +1,171 @@
...
+        const issueType = (e as Error).message.split("'")[1];
+        throw new SfdxError(`The specified metadata type is unsupported: [${issueType}]`);
+      } else {
+        throw e;
+      }
+    }
+
+    // This is only for debug output of matched files based on the command flags.
+    // It will log up to 20 file matches.
+    if (logger.debugEnabled && componentSet.size) {
```
shetzel, line 144: I've noticed that `logger.debugEnabled` doesn't seem to be working properly.  Probably a bug in sfdx-core.

shetzel (changes requested):

src/collections/componentSetBuilder.ts
```diff
@@ -0,0 +1,178 @@
...
+        const issueType = (e as Error).message.split("'")[1];
+        throw new SfdxError(`The specified metadata type is unsupported: [${issueType}]`);
+      } else {
+        throw e;
+      }
+    }
+
+    // This is only for debug output of matched files based on the command flags.
+    // It will log up to 20 file matches.
+    if (componentSet?.size) {
```
shetzel, line 151: sorry I wasn't clear with my comment. The `logger.debugEnabled` condition needs to be there so this doesn't execute unless debug is enabled.  I just meant it wasn't working in practice for some reason.

PR #579, opened 2022-02-15
Sm/file-responses-windows-paths-bug

mshanemc:
### What does this PR do?
SDR FileResponses were missing members on Windows for Documents (and probably other inFolder types) due to path separators not being handled when generating the message keys

the UT was building an invalid component (fullname includes folder when in a folder)

https://github.com/forcedotcom/cli/issues/1398
@W-10678360@

see github issue for expected/actual

QA:
You can repro with the user's repo/steps from the issue
You can also just use `force:source:deploy` and see the Document missing from the output


WillieRuemmele (approved):

shetzel:
### Test Results
✅ `sfdx force:source:beta:push`  twice - using project provided by customer
✅ `sfdx force:source:beta:push`  twice - using one my projects that has nested documents, reports, and email templates.

PR #583, opened 2022-02-28
Implement readFileSync for ZipTreeContainer

gbockus-sf:
### What does this PR do?
Implements the readFileSync function in ZipTreeContainer.  
I attempted to get this done without bringing in a third party module without success.  Basically if TS/js wants to be sync/async there's no magic formula for getting around it.  Also the zip module we are using `unzipper` has no support for reading a file synchronously. 

This PR brings in the deasync module that gets around this problem by providing a non-js solution for making an async flow synchronous.  

Required for fixing @W-10120433@

### Functionality Before

readFileSync was not implemented for ZipTreeContainer

### Functionality After

readFileSync is implemented for ZipTreeContainer.  Have verified that it fixes the issues found in https://github.com/forcedotcom/salesforcedx-vscode by linking the updated module locally.


METADATA_SUPPORT.md
```diff
@@ -483,15 +483,19 @@ v55 introduces the following new types.  Here's their current level of support
 
 |Metadata Type|Support|Notes|
 |:---|:---|:---|
```
gbockus-sf, line 485: Not sure how this magic happens.  commit flow I assume.  Lemme know if I need to fix up something here.

package.json
```diff
@@ -28,6 +28,7 @@
     "@salesforce/kit": "^1.5.32",
     "@salesforce/ts-types": "^1.4.2",
     "archiver": "^5.3.0",
```
gbockus-sf, line 30: This is the third party dependency that got the deal done. https://www.npmjs.com/package/deasync. MIT licensed and around 1M downloads a week so seemed a safe choice.  I'm following up on what I need to do for 3PP land.

src/resolve/treeContainers.ts
```diff
@@ -166,9 +167,23 @@ export class ZipTreeContainer extends TreeContainer {
...
+    this.readFile(fsPath)
+      .then((buffer) => {
+        dataBuffer = buffer;
+        done = true;
+      })
+      .catch((error) => {
+        done = true;
+        throw error;
+      });
+
```
gbockus-sf, line 182: This module covers doing exactly the ask here.  Do something async but treat the current flow and synchronous.

jag-sfdc (approved):

PR #595, opened 2022-03-21
Sm/parking-orbit

mshanemc:
requires https://github.com/forcedotcom/sfdx-core/pull/547 and a core bump to not break rest deploys

### What does this PR do?
breaking because
1. removed method on registryAccess for apiVersion
2. componentSet.getObject and componentSet.getPackageXml are now async
3. throws sfError instead of sfdxError

removes the hardcoded apiVersion from registry

behavior change: package.xml will always have a version.  If the componentSet isn't given one, SDR will make a callout to get the current version from the metadata coverage report.

this will break all the plugin-source NUTs

[@W-10809897@](https://gus.lightning.force.com/a07EE00000sIMg6YAG)


src/client/metadataTransfer.ts
```diff
@@ -160,7 +160,7 @@ export abstract class MetadataTransfer<Status extends MetadataRequestStatus, Res
         });
         if (target === 'source') {
           // for source convert the package.xml isn't included so write it separately
-          fs.writeFileSync(join(mdapiTempDir, 'package.xml'), source.getPackageXml());
+          fs.writeFileSync(join(mdapiTempDir, 'package.xml'), await source.getPackageXml());
```
WillieRuemmele, line 163: NIT: might as well `fs.promise.writeFile` 🤷‍♂️

WillieRuemmele (approved):
looks really good

src/convert/metadataConverter.ts
```diff
@@ -63,20 +63,22 @@ export class MetadataConverter {
...
-                tasks.push(promises.writeFile(destructiveManifestPath, destructiveManifestContents));
-              });
+              tasks.push(
+                destructiveChangesTypes.map(async (destructiveChangesType) =>
+                  promises.writeFile(
+                    join(packagePath, this.getDestructiveManifest(destructiveChangesType)),
+                    await cs.getPackageXml(4, destructiveChangesType)
+                  )
+                )
+              );
```
RodEsp, line 73: This doesn't seem right. Won't this cause `tasks` to end up with a nested array of `promises.writeFile` return types?
e.g. `tasks = [task1, task2, [promises.writeFile1, promises.writeFile2, promises.writeFile3], task3, task4, ...]`

Why are we even using `map` if the intent is just to iterate through the `destructiveChangesTypes` do some stuff and add them to `tasks`? Wouldn't `.forEach` be better and much clearer in intent?
```suggestion
              destructiveChangesTypes.forEach(async (destructiveChangesType) =>
                tasks.push(
                  promises.writeFile(
                    join(packagePath, this.getDestructiveManifest(destructiveChangesType)),
                    await cs.getPackageXml(4, destructiveChangesType)
                  )
                )
              );
```

src/convert/metadataConverter.ts
```diff
@@ -87,9 +89,9 @@ export class MetadataConverter {
             if (destructiveChangesTypes.length) {
               // for each of the destructive changes in the component set, convert and write the correct metadata
               // to each manifest
-              destructiveChangesTypes.map((destructiveChangeType) => {
+              destructiveChangesTypes.map(async (destructiveChangeType) => {
```
RodEsp, line 92: Why use `.map` instead of `.forEach` here?

RodEsp (changes requested):

src/convert/metadataConverter.ts
mshanemc: forEach doesn't work with async!

src/convert/metadataConverter.ts
mshanemc: `tasks` array is now typed so that mistake won't happen

src/convert/metadataConverter.ts
RodEsp: Oh... doi... you're right, I forgot.
How about just a `for ... of...` loop then?

src/convert/metadataConverter.ts
RodEsp: You're right about `.forEach` not working with `async` down below so how about a `for ... of` loop instead?

Example:
```suggestion
              for (const destructiveChangesType of destructiveChangesTypes) {
                tasks.push(
                  promises.writeFile(
                    join(packagePath, this.getDestructiveManifest(destructiveChangesType)),
                    await cs.getPackageXml(4, destructiveChangesType)
                  );
                );
              }
```

src/convert/metadataConverter.ts
WillieRuemmele: you bring up a good point. I honestly thought that `.forEach` and other array builtin methods were faster than a class `for` loop, but as I learned that's actually [backwards and incorrect](https://leanylabs.com/blog/js-forEach-map-reduce-vs-for-for_of/)

RodEsp: @WillieRuemmele I'm less concerned about performance and more concerned about readability & maintainability of the code. 

`.map` creates a new array based on the transformations you apply in its callback, if all you want to do is iterate over an array and you're not going to do anything with the new array `.map` creates then better to just use a loop, regardless of performance considerations (we're not doing anything here that requires millisecond improvements). Using methods that aren't intended for what you're trying to do makes it harder for developers to understand what the code _should_ do and leads to bugs like [this one](https://github.com/forcedotcom/source-deploy-retrieve/pull/595#discussion_r838856708).

mshanemc: I really intend to return a new array of Promises that gets added to `tasks`.  Having a type on `tasks` would have also prevented the bug.

src/convert/metadataConverter.ts
```diff
@@ -63,20 +63,22 @@ export class MetadataConverter {
...
-                tasks.push(promises.writeFile(destructiveManifestPath, destructiveManifestContents));
-              });
+              tasks.push(
+                ...destructiveChangesTypes.map(async (destructiveChangesType) =>
+                  promises.writeFile(
+                    join(packagePath, this.getDestructiveManifest(destructiveChangesType)),
+                    await cs.getPackageXml(4, destructiveChangesType)
+                  )
+                )
+              );
```
RodEsp, line 73: I'd still rather see this use a `for ... of` loop. It's much more straight forward than using `.map` + array destructuring and accomplishes the same task.

```suggestion
              for (const destructiveChangesType of destructiveChangesTypes) {
                tasks.push(
                  promises.writeFile(
                    join(packagePath, this.getDestructiveManifest(destructiveChangesType)),
                    await cs.getPackageXml(4, destructiveChangesType)
                  );
                );
              }
```

RodEsp (changes requested):

RodEsp (approved):

PR #601, opened 2022-03-31
fix: resolve strictDirectoryName types in mdapi format

shetzel:
### What does this PR do?
Fixes `MetadataResolver` to match a file path in source format and mdapi format when testing types with strictDirectoryNames that also do not have an adapter strategy defined (those types are resolved properly another way).

This should fix the following types:
- CustomSite
- IntegrationHubSettings
- AppointmentSchedulingPolicy
- RestrictionRule
- ObjectHierarchyRelationship
- IndustriesManufacturingSettings
- DataSource
- FieldRestrictionRule
- AppointmentAssignmentPolicy

https://github.com/forcedotcom/cli/issues/1448
@W-10910004@

### Functionality Before
Would not resolve files in mdapi format with certain strictDirectoryName types such as CustomSite.  This would cause (e.g.) a retrieval of these types to fail to be written to the file system.

### Functionality After
Files resolve properly.  Metadata retrieval of these files are successful.

### DIFF
--- a/src/resolve/metadataResolver.ts
+++ b/src/resolve/metadataResolver.ts
@@ -174,8 +174,8 @@ export class MetadataResolver {
           // any of the following 3 options is considered a good match
           // mixedContent and bundles don't have a suffix to match
           ['mixedContent', 'bundle'].includes(type.strategies?.adapter) ||
-          // the suffix matches the type we think it is
-          (type.suffix && fsPath.endsWith(`${type.suffix}${META_XML_SUFFIX}`)) ||
+          // the file suffix (in source or mdapi format) matches the type suffix we think it is
+          (type.suffix && [type.suffix, `${type.suffix}${META_XML_SUFFIX}`].some((s) => fsPath.endsWith(s))) ||
           // the type has children and the path also includes THAT directory
           (type.children?.types &&
             Object.values(type.children?.types)
--- a/test/resolve/metadataResolver.test.ts
+++ b/test/resolve/metadataResolver.test.ts
@@ -182,6 +182,21 @@ describe('MetadataResolver', () => {
         expect(mdResolver.getComponentsFromPath(path)).to.deep.equal([expectedComponent]);
       });
 
+      it('Should determine type for EmailServicesFunction metadata file (mdapi format)', () => {
+        const path = join('unpackaged', 'emailservices', 'MyEmailServices.xml');
+        const treeContainer = VirtualTreeContainer.fromFilePaths([path]);
+        const mdResolver = new MetadataResolver(undefined, treeContainer);
+        const expectedComponent = new SourceComponent(
+          {
+            name: 'MyEmailServices',
+            type: registry.types.emailservicesfunction,
+            xml: path,
+          },
+          treeContainer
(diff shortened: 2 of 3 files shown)


test/resolve/metadataResolver.test.ts
```diff
@@ -142,6 +142,46 @@ describe('MetadataResolver', () => {
         expect(access.getComponentsFromPath(path)).to.deep.equal([matchingContentFile.COMPONENT]);
       });
 
+      it('Should determine type for metadata file with known suffix and strictDirectoryName', () => {
```
shetzel, line 145: I made sure the test failed without the fix.

src/resolve/metadataResolver.ts
```diff
@@ -174,8 +174,8 @@ export class MetadataResolver {
           // any of the following 3 options is considered a good match
           // mixedContent and bundles don't have a suffix to match
           ['mixedContent', 'bundle'].includes(type.strategies?.adapter) ||
-          // the suffix matches the type we think it is
-          (type.suffix && fsPath.endsWith(`${type.suffix}${META_XML_SUFFIX}`)) ||
+          // the file suffix (in source or mdapi format) matches the type suffix we think it is
+          (type.suffix && [type.suffix, `${type.suffix}${META_XML_SUFFIX}`].some((s) => fsPath.endsWith(s))) ||
```
mshanemc, line 178: that exception is that little punk emailServicesFunction for whom `.xml` is the suffix.

now that we're testing with a real registry, can you make sure that it's handled correctly?

mshanemc (changes requested):
let's test with emailServicesFunction

WillieRuemmele (approved):

mshanemc (approved):

src/resolve/metadataResolver.ts
mshanemc: FYI, I was wanting it in a strictDir.  ESF has a habit of matching way too promiscuously and I'm trying to keep that from happening.

https://github.com/forcedotcom/source-deploy-retrieve/pull/595/files#diff-af7a780a5aa66c4831707c3ccd4e24d4f88a6639346bc0b1897c287efcbe14c9

So...what would happen if it *were* in strictDir?

shetzel: Testing would help us out, but I think it would still work just fine because it would only get to the suffix matching if the path contained the expected directory.

shetzel: After testing, ESF still retrieves and deploys successfully when defined as a strictDirectoryName.

WillieRuemmele:
QA notes: 
✅ Using `CustomSite` it worked for `retrieve`
image
✅ : `source:convert` -> `mdapi:beta:deploy` 
✅ : local change, picked up `source:status`
✅ : `source:push`

WillieRuemmele:
the [release notes](https://github.com/forcedotcom/cli/blob/main/releasenotes/sfdx/README.md#salesforce-cli-release-notes-sfdx-commands) explain how to get the RC version `sfdx update stable-rc`

PR #606, opened 2022-04-04
fix: add telemetry to metadata api deploy

RodEsp:
### What does this PR do?

Enables sending telemetry data to App Insights for mdapi deploys

@W-10746359@

Replaces #596

### DIFF
--- a/src/collections/componentSet.ts
+++ b/src/collections/componentSet.ts
@@ -5,8 +5,11 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 /* eslint  @typescript-eslint/unified-signatures:0 */
+
 import { j2xParser } from 'fast-xml-parser';
-import { AuthInfo, Connection, Logger } from '@salesforce/core';
+
+import { AuthInfo, Connection, Lifecycle, Logger } from '@salesforce/core';
+
 import {
   MetadataApiDeploy,
   MetadataApiDeployOptions,
@@ -15,26 +18,26 @@ import {
 } from '../client';
 import { XML_DECL, XML_NS_KEY, XML_NS_URL } from '../common';
 import { ComponentSetError } from '../errors';
+import { MetadataType, RegistryAccess } from '../registry';
 import {
   ComponentLike,
+  ConnectionResolver,
   ManifestResolver,
   MetadataComponent,
   MetadataMember,
   MetadataResolver,
-  ConnectionResolver,
   SourceComponent,
   TreeContainer,
 } from '../resolve';
-import { MetadataType, RegistryAccess } from '../registry';
+import { LazyCollection } from './lazyCollection';
 import {
   DestructiveChangesType,
+  FromConnectionOptions,
   FromManifestOptions,
   FromSourceOptions,
-  FromConnectionOptions,
   PackageManifestObject,
   PackageTypeMembers,
 } from './types';
-import { LazyCollection } from './lazyCollection';
 
 export type DeploySetOptions = Omit<MetadataApiDeployOptions, 'components'>;
 export type RetrieveSetOptions = Omit<MetadataApiRetrieveOptions, 'components'>;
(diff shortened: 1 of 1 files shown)


RodEsp:
I dunno what's up with that one failing unit test. I can't figure it out but it's also failing in the same way on the `main` branch so I don't think it has to do with my changes.

src/collections/componentSet.ts
```diff
@@ -292,6 +295,20 @@ export class ComponentSet extends LazyCollection<MetadataComponent> {
...
+    const listOfMetadataTypesDeployed = Array.from(new Set(toDeploy.map((c) => c.type.name)));
+
+    void Lifecycle.getInstance().emitTelemetry({
+      eventName: 'metadata_api_deploy',
+      library: 'SDR',
+      function: 'ComponentSet.deploy',
+      totalNumberOfComponents: this.size,
+      numberOfComponentsDeployed: toDeploy.length,
+      componentsDeployed: listOfMetadataTypesDeployed.toString(),
+      componentsDeployedTruncated: listOfMetadataTypesDeployed.length < 8000 ? false : true,
```
peternhale, line 309: Just curious about the number 8000. Is this a limit known to the customer?

peternhale (dismissed):

src/collections/componentSet.ts
shetzel: That number is for unique metadata types in the deploy request.  Salesforce has around 350 metadata types so not sure where 8000 comes from.  Also, what's being truncated?

src/collections/componentSet.ts
```diff
@@ -292,6 +295,20 @@ export class ComponentSet extends LazyCollection<MetadataComponent> {
...
     await mdapiDeploy.start();
+
+    // Creates an array of unique metadata types to be deployed, uses Set to avoid duplicates.
+    const listOfMetadataTypesDeployed = Array.from(new Set(toDeploy.map((c) => c.type.name)));
+
+    void Lifecycle.getInstance().emitTelemetry({
+      eventName: 'metadata_api_deploy',
+      library: 'SDR',
+      function: 'ComponentSet.deploy',
+      totalNumberOfComponents: this.size,
```
shetzel, line 306: How is `totalNumberOfComponents` different in a significant way from `numberOfComponentsDeployed` at this point in the function?

src/collections/componentSet.ts
```diff
@@ -292,6 +295,20 @@ export class ComponentSet extends LazyCollection<MetadataComponent> {
...
+      numberOfComponentsDeployed: toDeploy.length,
+      componentsDeployed: listOfMetadataTypesDeployed.toString(),
```
shetzel, line 308: We should be careful about the names of these fields.  Nothing has been deployed yet.  Everything at this point is, "to be deployed", or "the metadata being requested for deployment".
`componentsDeployed` as a name can't be correct since you're providing the unique metadata types, not the components.

src/collections/componentSet.ts
RodEsp: @shetzel not entire sure, I just copied this from the PR where this telemetry was removed from toolbelt.
https://github.com/salesforcecli/toolbelt/pull/215/files#diff-ca446f5ccf231e22c85434264adcb8d977f76d6375c1ae789fd39c1bde5aa324L330

src/collections/componentSet.ts
RodEsp: @shetzel I'm not entirely sure. The [original telemetry](https://github.com/salesforcecli/toolbelt/pull/215/files#diff-ca446f5ccf231e22c85434264adcb8d977f76d6375c1ae789fd39c1bde5aa324L313) used packages but since the concept of packages doesn't really exist in SDR I changed it to components.

I used `toDeploy` for `numberOfComponentsDeployed` because the variable name implies that's what being deployed and I couldn't figure out another way to identify that. If that's not the case though I can remove it and just leave `totalNumberOfComponents`.

If that's not the case and both of these numbers are the same I can just remove one.

src/collections/componentSet.ts
RodEsp: I'll change it.

@shetzel I pulled this naming directly from the PR where this was removed from toolbelt.
https://github.com/salesforcecli/toolbelt/pull/215/files#diff-ca446f5ccf231e22c85434264adcb8d977f76d6375c1ae789fd39c1bde5aa324L329

src/collections/componentSet.ts
shetzel: That original telemetry (iirc) was primarily added to gain some insight about how many customers/deploys were done as MPD (multiple package directory), and how big the projects were (number of types).  That 8000 number looks like it was originally for individual metadata but the code was changed to be about metadata types, otherwise it doesn't make sense.

I think we should change this a bit to be more useful.  Let's chat about it.

RodEsp:
@shetzel I just updated this based on what you said in the quip doc. We should now get telemetry for deploy results too. Lemme know what you think.

shetzel (approved):

shetzel:
I'm merging this despite a failing, flapping NUT since the test has passed in multiple previous runs and the changes are not in retrieve.

PR #614, opened 2022-04-21
Sm/timeout id in sfdx error

mshanemc:
What does this PR do?
throws an sfdxError with data property for id on timeout.

I'm making a separate WI for SDR to use error/messages from core rather than its own framework.

[@W-10978703@](https://gus.lightning.force.com/a07EE00000ujyy4YAA)

for QA...SDR is emitting error.code 1 for backward compatibility.  I think the logic on plugin-source should set incompletes to 69 based on the mdapi status...but that's worth verifying


shetzel (approved):

shetzel:
The failing NUTs passed locally when I ran them.  Manual testing looked good as well.

PR #615, opened 2022-04-22
fix: resolveTypeFromStrictFolder for children of decomposed component types with decomposition topLevel

jayree:
### What does this PR do?

Fix function `resolveTypeFromStrictFolder` for children of decomposed component types with decomposition `topLevel` e.g. (`customobjecttranslation`, `bot`)

(I'm unsure about `dataMappingObjectDefinitions` because I can't test it.)

@W-0@

### Functionality Before

`resolver.getComponentsFromPath` doesn't resolve parent component for child types e.g. `customfieldtranslation`, `botversion`

### Functionality After

`resolver.getComponentsFromPath` resolves parent component correctly for child types e.g. `customfieldtranslation`, `botversion`

The current solution doesn't match with decomposed component types with decomposition `topLevel` as the `childType.directoryName` is not part of the file path `fspath`.

### DIFF
--- a/src/resolve/metadataResolver.ts
+++ b/src/resolve/metadataResolver.ts
@@ -176,11 +176,11 @@ export class MetadataResolver {
           ['mixedContent', 'bundle'].includes(type.strategies?.adapter) ||
           // the file suffix (in source or mdapi format) matches the type suffix we think it is
           (type.suffix && [type.suffix, `${type.suffix}${META_XML_SUFFIX}`].some((s) => fsPath.endsWith(s))) ||
-          // the type has children and the path also includes THAT directory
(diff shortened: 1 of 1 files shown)


jayree:
@shetzel @mshanemc this bug was introduced in `v5.1.3`:

Example:
```typescript
import {MetadataResolver} from '@salesforce/source-deploy-retrieve';
const resolver = new MetadataResolver();
const [component] = resolver.getComponentsFromPath('force-app/main/default/objectTranslations/Account-en_US/account_name_local.fieldTranslation-meta.xml');

console.log(component.parent?.fullName);
// expected: Account-en_US
// actual: undefined
```

PR #618, opened 2022-04-26
Sm/nested-lwc

mshanemc:
### What does this PR do?
restore something that worked under toolbelt ex: /lwc/folder/lwc/cmp
NUTs will be added to plugin-source for this

https://github.com/forcedotcom/cli/issues/1477
@W-11051996@

QA notes: try this with a retrieve because I haven't.


shetzel (approved):

shetzel:
I tested push, deploy, and retrieve with an LWC in the path: `force-app/lwc/foo/lwc/myLWC/**` and everything looked good.  I also tested with the plugin-source NUTs and they passed as well.

PR #644, opened 2022-06-06
Sm/sdr-events

mshanemc:
### What does this PR do?
support for event-based source tracking
- emits new Lifecycle events pre/post-deploy/retrieve

[@W-11186378@](https://gus.lightning.force.com/a07EE00000xc1DjYAI)


src/client/metadataApiDeploy.ts
```diff
@@ -348,11 +349,17 @@ export class MetadataApiDeploy extends MetadataTransfer<MetadataApiDeployStatus,
...
-    ]);
+    const connection = await this.getConnection();
+    // store for use in the scopedPostDeploy event
+    this.orgId = connection.getAuthInfoFields().orgId;
+    // only do event hooks if source, (NOT a metadata format) deploy
+    if (this.options.components) {
+      await Lifecycle.getInstance().emit('scopedPreDeploy', {
+        componentSet: this.options.components,
+        orgId: this.orgId,
+      } as ScopedPreDeploy);
```
shetzel, line 360: Is this meant to replace the existing predeploy hooks in plugin-source or will they coincide for the life of sfdx?  Is this an internal only hook or can other plugins listen?

shetzel (approved):

src/client/metadataApiDeploy.ts
mshanemc: Mike and I left that question TBD. We could either make these the official hooks, or maybe even emit the current sfdx hooks from SDR too.

There's not really a sense of "privileged" hooks, so there's nothing to stop people from seeing and using them.

PR #652, opened 2022-06-21
Add retry to StandardValueSet query to fix network errors

iowillhoit:
### What does this PR do?
When running `force:source:manifest:create --fromorg`, the `singleRecordQuery` method to retrieve `StandardValueSet` was having connection errors, added retry logic.

[@W-11303606@](https://gus.my.salesforce.com/apex/ADM_WorkLocator?bugorworknumber=W-11303606)

### Functionality Before
Connection errors would cause some queries to fail. The result was different package.xml file almos every time you ran `force:source:manifest:create --fromorg YOUR_ORG`

https://user-images.githubusercontent.com/1715111/174886528-5f1d222f-d61d-4616-9b41-dc3ab24c6e4c.mov

### Functionality After
Failed queries retry, the package.xml is the same every time

### Testing 
- Pull and cd into this `source-deploy-retreive` branch
- `yarn build`
- `yarn link` (if needed)
- Pull and cd into `plugin-source`
- `yarn link @salesforce/source-deploy-retrieve`
- `sfdx plugins:link .`
- Create a new Dreamhouse org
  - `git clone git@github.com:trailheadapps/dreamhouse-lwc.git dreamhouse-testing-standard-value-set`
  - `cd dreamhouse-testing-standard-value-set`
  - `sfdx force:org:create -s -d 1 -f config/project-scratch-def.json`
  - Get the username
- Run `sfdx force:source:manifest:create --fromorg THE_USERNAME`
- Commit the created `package.xml`
- Run the command several more times and use `git diff` to confirm not changes occur over multiple runs
  - Note: you can see the retries by adding the following here
```js
,
{
  logger: (msg) => {
    console.log(msg);
  },
}
```

# Now unlink everything! 🔥🔗

### DIFF
--- a/src/resolve/connectionResolver.ts
+++ b/src/resolve/connectionResolver.ts
@@ -6,6 +6,7 @@
  */
 
 import { Connection, Logger } from '@salesforce/core';
+import { retry, NotRetryableError, RetryError } from 'ts-retry-promise';
 import { RegistryAccess, registry as defaultRegistry, MetadataType } from '../registry';
 import { standardValueSet } from '../registry/standardvalueset';
 import { FileProperties, StdValueSetRecord, ListMetadataQuery } from '../client/types';
@@ -108,10 +109,32 @@ export class ConnectionResolver {
     if (query.type === defaultRegistry.types.standardvalueset.name && members.length === 0) {
       const standardValueSetPromises = standardValueSet.fullnames.map(async (standardValueSetFullName) => {
         try {
-          const standardValueSetRecord: StdValueSetRecord = await this.connection.singleRecordQuery(
-            `SELECT Id, MasterLabel, Metadata FROM StandardValueSet WHERE MasterLabel = '${standardValueSetFullName}'`,
-            { tooling: true }
-          );
+          // The 'singleRecordQuery' method was having connection errors, using `retry` resolves this
+          // Note that this type of connection retry logic may someday be added to jsforce v2
+          // Once that happens this logic could be reverted
+          const standardValueSetRecord: StdValueSetRecord = await retry(async () => {
+            let result: StdValueSetRecord;
+
+            try {
+              result = await this.connection.singleRecordQuery(
+                `SELECT Id, MasterLabel, Metadata FROM StandardValueSet WHERE MasterLabel = '${standardValueSetFullName}'`,
+                { tooling: true }
+              );
+            } catch (err) {
+              // We exit the retry loop with `NotRetryableError` if we get an (expected) unsupported metadata type error
+              const error = err as Error;
+              if (error.message.includes('either inaccessible or not supported in Metadata API')) {
+                this.logger.debug('Expected error:', error.message);
+                throw new NotRetryableError(error.message);
+              }
+
+              // Otherwise throw the err so we can retry again
+              throw err;
+            }
+
+            return result;
+          });
+
           return (
             standardValueSetRecord.Metadata.standardValue.length && {
               fullName: standardValueSetRecord.MasterLabel,
@@ -126,8 +149,15 @@ export class ConnectionResolver {
               lastModifiedDate: '',
             }
           );
-        } catch (error) {
-          this.logger.debug((error as Error).message);
+        } catch (err) {
+          // error.message here will be overwritten by 'ts-retry-promise'
+          // Example error.message from the library: "All retries failed" or "Met not retryable error"
+          // 'ts-retry-promise' exposes the actual error on `error.lastError`
+          const error = err as RetryError;
+
+          if (error.lastError && error.lastError.message) {
+            this.logger.debug(error.lastError.message);
+          }
         }
       });
       for await (const standardValueSetResult of standardValueSetPromises) {


src/resolve/connectionResolver.ts
```diff
@@ -108,10 +109,32 @@ export class ConnectionResolver {
...
       const standardValueSetPromises = standardValueSet.fullnames.map(async (standardValueSetFullName) => {
         try {
-          const standardValueSetRecord: StdValueSetRecord = await this.connection.singleRecordQuery(
-            `SELECT Id, MasterLabel, Metadata FROM StandardValueSet WHERE MasterLabel = '${standardValueSetFullName}'`,
-            { tooling: true }
-          );
+          // The 'singleRecordQuery' method was having connection errors, using `retry` resolves this
+          // Note that this type of connection retry logic may someday be added to jsforce v2
+          // Once that happens this logic could be reverted
+          const standardValueSetRecord: StdValueSetRecord = await retry(async () => {
```
peternhale, line 115: @iowillhoit why are there no limits being applied to the retry? The `retry` function defaults to retries and timeout of [INFINITELY](https://github.com/normartin/ts-retry-promise#configuration).

peternhale (changes requested):

src/resolve/connectionResolver.ts
iowillhoit: @peternhale It looks like the `number | INFINITELY` in the docs are Types. The defaults are `10` retries and a `60s` for timeouts.

.circleci/config.yml
```diff
@@ -93,14 +93,20 @@ jobs:
             equal: ['linux', <<parameters.os>>]
           steps:
             - run: yarn add $CIRCLE_PROJECT_USERNAME/$CIRCLE_PROJECT_REPONAME#$CIRCLE_SHA1
+      - run:
+          # STL could also have an SDR inside it.  Take it out to prevent conflicts
+          name: remove sdr from source-tracking
+          command: |
+            npm install shx -g
+            shx rm -rf node_modules/@salesforce/source-deploy-retrieve
+          working_directory: node_modules/@salesforce/source-tracking
```
gbockus-sf, line 102: Is this a common thing to do for circular deps? 
Does this not mean that for the tests source-tracking would be using the current version of SDR, but after deploy it'd be using whatever version it specifies in it's package.json?

src/resolve/connectionResolver.ts
```diff
@@ -126,8 +149,15 @@ export class ConnectionResolver {
...
-          this.logger.debug((error as Error).message);
+        } catch (err) {
+          // error.message here will be overwritten by 'ts-retry-promise'
+          // Example error.message from the library: "All retries failed" or "Met not retryable error"
+          // 'ts-retry-promise' exposes the actual error on `error.lastError`
+          const error = err as RetryError;
+
+          if (error.lastError && error.lastError.message) {
+            this.logger.debug(error.lastError.message);
+          }
```
gbockus-sf, line 160: Should we add a few unit tests to cover the conditions added here? 
- retries and succeeds 
- Hits the retry limit (seemed a bit like exercising the retry lib, but would be good to have a test to verify what happens when it hits the limit). Could just stub the failure condition early verse having to actually call until the lib fails. 
- Not Retryable error thrown

It would make the unit testing a bit cleaner if you refactored the `standardValueSetRecord` specific handling out into a method

yarn.lock
```diff
@@ -6436,7 +6436,7 @@ typedoc@0.22.17:
     minimatch "^5.1.0"
     shiki "^0.10.1"
 
-typescript@^4.1.3, typescript@^4.4.3:
+typescript@^4.1.3, typescript@^4.4.3, typescript@^4.7.4:
```
gbockus-sf, line 6439: Seems like several of the deps are now out of sync with what's in `@salesforce/dev-scripts`.  Just verifying that's the way to go.

jayree:
@iowillhoit I know these problems from the past. However, I have not seen this issue since api version 50. What error messages appear in the debug log?

.circleci/config.yml
mshanemc: they're not circular, just transitive.  plugin-source depends on SDR and STL and STL depends on SDR.

When we build the CLI, we bump both SDR and STL via yarn resolutions to their latest version.

This tests the changes on this branch everywhere that SDR is used.  Imagine a bug in SDR that causes problems for STL, but you don't catch it because the nuts are using an STL with a different (bug-free) SDR.  Wouldn't want to miss that!

src/resolve/connectionResolver.ts
```diff
@@ -108,10 +109,32 @@ export class ConnectionResolver {
...
+              if (error.message.includes('either inaccessible or not supported in Metadata API')) {
+                this.logger.debug('Expected error:', error.message);
+                throw new NotRetryableError(error.message);
+              }
+
+              // Otherwise throw the err so we can retry again
+              throw err;
+            }
+
+            return result;
```
mshanemc, line 135: if you return the result from the try {} block, that would avoid the `let` , the assignment, and possibly the `await`

iowillhoit:
> @iowillhoit I know these problems from the past. However, I have not seen this issue since api version 50. What error messages appear in the debug log?

@jayree Here are the two types of errors I am seeing:
TLS Connection:
```
request to https://customer-inspiration-1104-dev-ed.my.salesforce.com/services/data/v55.0/tooling/query?q=SELECT%20Id%2C%20MasterLabel%2C%20Metadata%20FROM%20StandardValueSet%20WHERE%20MasterLabel%20%3D%20%27ProcessExceptionStatus%27 failed, reason: Client network socket disconnected before secure TLS connection was established
```
Connection Reset:
```
FetchError: request to https://customer-inspiration-1104-dev-ed.my.salesforce.com/services/data/v55.0/tooling/query?q=SELECT%20Id%2C%20MasterLabel%2C%20Metadata%20FROM%20StandardValueSet%20WHERE%20MasterLabel%20%3D%20%27WorkTypeGroupAddInfo%27 failed, reason: read ECONNRESET
    at ClientRequest.<anonymous> (/Users/ewillhoit/dev/source-deploy-retrieve/node_modules/node-fetch/lib/index.js:1491:11)
```

src/resolve/connectionResolver.ts
iowillhoit: Added some tests @gbockus-sf 👍

WillieRuemmele (approved):

peternhale (approved):

WillieRuemmele:
### QA NOTES:
---
✅ : following instructions in post, generated manifest 5 times, no differences
✅ : sanity check `--metadata StandardValueSet` 5 times, no differences
✅ : unlink SDR/PS see varying results in `package.xml`

PR #663, opened 2022-07-12
fix: static resources without content due to ignore produce nice error

mshanemc:
### What does this PR do?
when forceignore causes *all* of the content in a static resource to be ignored, `tree.exists(undefied)` is called which makes an uglier error.

now it produces a helpful error about which component is causing the problem.

and has a UT for the behavior

@W-11148485@
https://github.com/forcedotcom/cli/issues/1516


shetzel (approved):

shetzel:
I see this error message with the fixes:
`ERROR running force:source:push:  force-app/main/default/staticresources/testResource.resource-meta.xml: Expected source files for type 'StaticResource'`. 👍

PR #666, opened 2022-07-14
fix: only scan pkgDirs when merging non-decomposed MD

cristiand391:
### What does this PR do?

Updates `NonDecompositionFinalizer` to only scan the package directories in a DX project when merging non-decomposed metadata.
If the `defaultDirectory` passed to `finalize` isn't a package dir it will ignore it.

See: https://developer.salesforce.com/docs/atlas.en-us.sfdx_dev.meta/sfdx_dev/sfdx_dev_ws_mpd.htm

https://github.com/forcedotcom/cli/issues/1540
@W-11186731@

### Functionality Before

`NonDecompositionFinalizer` scans components in root project dir which causes issues when components are found in other folders (mdapi, etc).

### Functionality After

`NonDecompositionFinalizer` will only scan components found inside of package directories, if the finalizer gets a defaultDirectory that's not a package it will threat it as the target output dir for a transformation and not scan it.

### DIFF
--- a/src/convert/convertContext.ts
+++ b/src/convert/convertContext.ts
@@ -4,8 +4,9 @@
  * Licensed under the BSD 3-Clause license.
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
-import { dirname, join, resolve } from 'path';
+import { join } from 'path';
 import { getString, JsonArray, JsonMap } from '@salesforce/ts-types';
+import { SfProject } from '@salesforce/core';
 import { META_XML_SUFFIX, XML_NS_KEY, XML_NS_URL } from '../common';
 import { ComponentSet } from '../collections';
 import { normalizeToArray } from '../utils';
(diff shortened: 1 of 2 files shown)


src/convert/convertContext.ts
```diff
@@ -193,12 +194,21 @@ class NonDecompositionFinalizer extends ConvertTransactionFinalizer<NonDecomposi
...
-      this.transactionState.exampleComponent.type.name
-    );
+    let allNonDecomposed: SourceComponent[];
+
+    if (pkgPaths.includes(defaultDirectory)) {
+      allNonDecomposed = this.getAllComponentsOfType(pkgPaths, this.transactionState.exampleComponent.type.name);
+    } else {
+      // defaultDirectory isn't a package, assumes it's the target output dir for conversion
+      // so no need to scan this folder
+      allNonDecomposed = [];
```
cristiand391, line 210: This is the output dir passed to the converter in the conversion pipeline, if the dir isn't a package in a project SDR don't need to scan it.

https://github.com/forcedotcom/source-deploy-retrieve/blob/3af34cedbccb2725e934eceda82331798030ba82/src/convert/metadataConverter.ts#L116

When running `sfdx force:mdapi:convert -r mdapi -d src` SDR will not look for non-decomposed metadata in `/home/user/dreamhouse-lwc/src`.

WillieRuemmele (approved):

WillieRuemmele:
## QA Notes
---
✅ : issue repository/steps now `mdapi:convert`s `CustomLabels`
✅ : ignores MPD CLs correctly

PR #671, opened 2022-07-28
Removing support for Connect Center

elisfdc:
### What does this PR do?
Removes entities used by the Connect Center product. 
Feature is no longer in the pipeline
@W-11488939@


peternhale (approved):

elisfdc:
@peternhale , who can review for "forcedotcom/pdt"

package.json
```diff
@@ -112,7 +112,7 @@
     ]
   },
   "volta": {
-    "node": "12.11.0",
+    "node": "16.16.0",
```
gbockus-sf, line 115: Should we bump the engines property here as well?

elisfdc: This should probably match the engines property but I changed it to match what I had. Seems like this would have been failing anyone previously using volta.

elisfdc: Up to you guys on what this value should be. I can make the change

gbockus-sf: yeah I use (and think mostly our team does) nvm and went to verify there wasn't a .nvmrc as well for the same reason.  My local env currently defaults to v14.

gbockus-sf (approved):
Should be fine for PDT 👍

peternhale (approved):

mshanemc:
the sister PR was merged, so closing this one.

PR #684, opened 2022-08-04
Sm/max-api-version-on-deploys

mshanemc:
### What does this PR do?

1. when the componentSet has made it to the SDR deploy step without any apiVersion info, set it from what's on the connection so it doesn't default to highest per mdapi coverage in the package.xml being sent
2. getConnection will use org's max, emiting a warning if the user is trying to use something too high, in case the default isn't available yet to avoid issues on other endpoints polling/cancel/etc

[NB the proper way to do this is to set `sourceApiVersion` in your `sfdx-project.json`.  This is the workaround for tooling to use more intelligent default behavior.

@W-11549415@
https://github.com/forcedotcom/cli/issues/1656

### Functionality Before

```
=== Component Failures [1]

 Type  Name        Problem                        
 ───── ─────────── ────────────────────────────── 
 Error package.xml Invalid version specified:56.0 
```

### Functionality After

pushes as expected

### DIFF
--- a/src/client/metadataApiDeploy.ts
+++ b/src/client/metadataApiDeploy.ts
@@ -352,6 +352,11 @@ export class MetadataApiDeploy extends MetadataTransfer<MetadataApiDeployStatus,
     const connection = await this.getConnection();
     // store for use in the scopedPostDeploy event
     this.orgId = connection.getAuthInfoFields().orgId;
+    if (this.components && !this.components.apiVersion && !this.components.sourceApiVersion) {
+      // if we have a ComponentSet, but got no version info, let's use the org's max version for calculating what goes into the package.xml
+      this.components.apiVersion = connection.getApiVersion();
+      this.components.sourceApiVersion = connection.getApiVersion();
+    }
     // only do event hooks if source, (NOT a metadata format) deploy
     if (this.options.components) {
       await Lifecycle.getInstance().emit('scopedPreDeploy', {
--- a/src/client/metadataTransfer.ts
+++ b/src/client/metadataTransfer.ts
@@ -192,7 +192,7 @@ export abstract class MetadataTransfer<Status extends MetadataRequestStatus, Res
         this.logger.debug(`Overriding apiVersion to: ${this.apiVersion}`);
       }
     }
-    return this.usernameOrConnection;
+    return getConnectionNoHigherThanOrgAllows(this.usernameOrConnection, this.apiVersion);
   }
 
   private async poll(): Promise<StatusResult> {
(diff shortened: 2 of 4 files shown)


src/client/metadataApiDeploy.ts
```diff
@@ -352,6 +352,11 @@ export class MetadataApiDeploy extends MetadataTransfer<MetadataApiDeployStatus,
     const connection = await this.getConnection();
     // store for use in the scopedPostDeploy event
     this.orgId = connection.getAuthInfoFields().orgId;
+    if (this.components && !this.components.apiVersion && !this.components.sourceApiVersion) {
```
peternhale, line 355: @mshanemc I am not too sure about this assertion, but looking at component set class there is a function `getObject` that [establishes apiVersion](https://github.com/forcedotcom/source-deploy-retrieve/blob/sm/max-api-version-on-deploys/src/collections/componentSet.ts#L327) for the component. Is it possible that `getObject` has already been called before reaching this check?

peternhale (changes requested):

src/client/metadataApiDeploy.ts
mshanemc: Within SDR, that's called during the creation of the package.xml, which is pretty late.

Technically, some consumer could do anything they want with it, but it's unlikely.

mshanemc: what change are you requesting?

peternhale: Nothing to change if my suspicion was not confirmed. Thanks

peternhale (approved):

src/client/metadataTransfer.ts
```diff
@@ -269,3 +269,16 @@ export abstract class MetadataTransfer<Status extends MetadataRequestStatus, Res
   protected abstract pre(): Promise<AsyncResult>;
   protected abstract post(result: Status): Promise<Result>;
 }
+
+/* prevent requests on apiVersions higher than the org supports */
+const getConnectionNoHigherThanOrgAllows = async (conn: Connection, requestedVersion: string): Promise<Connection> => {
```
gbockus-sf, line 274: Do we have test coverage for this method?

src/client/metadataApiDeploy.ts
```diff
@@ -352,6 +352,11 @@ export class MetadataApiDeploy extends MetadataTransfer<MetadataApiDeployStatus,
...
+      // if we have a ComponentSet, but got no version info, let's use the org's max version for calculating what goes into the package.xml
```
gbockus-sf, line 356: thanks for this comment. def adds value.

src/client/metadataTransfer.ts
```diff
@@ -269,3 +269,16 @@ export abstract class MetadataTransfer<Status extends MetadataRequestStatus, Res
...
+  // uses a TTL cache, so mostly won't hit the server
+  const maxApiVersion = await conn.retrieveMaxApiVersion();
+  if (requestedVersion && parseInt(requestedVersion, 10) > parseInt(maxApiVersion, 10)) {
```
gbockus-sf, line 277: Would it be worthwhile to validate the requested version here?  It seems the caller could send an invalid version string here and get by without altering the max version allowed on the connection.

gbockus-sf (approved):
a few question in there but change looks good.

shetzel (approved):

src/client/metadataTransfer.ts
mshanemc: it's being executed on every deploy, so about 100x.
At least some of those specify an apiVersion...not sure if any don't and would check the if conditional.

src/client/metadataTransfer.ts
mshanemc: the ComponentSet allows consumers to set its apiVersion and sourceApiVersion properties.  If there's a place to put validation, it'd be there.  This is internal non-exported function, only called from this file, which is extended by Deploy and Retrieve.

Are you wanting to put getters/setters so we can do validation on those props for ComponentSet?

PR #695, opened 2022-08-26
chore: Support DigitalExperienceBundle, DigitalExperience and DigitalExperienceConfig MD types in SDR

Kaaviyan:
### What does this PR do?
Support DigitalExperienceBundle, DigitalExperience, and DigitalExperienceConfig MD types in SDR (without support for source tracking commands)

@W-11492368@

### Functionality Before

1. sfdx force:source and force:mdapi commands didn't support DigitalExperienceBundle, DigitalExperience, and DigitalExperienceConfig MD types 
2. Partial wildcards retrievals for child types were not supported 

### Functionality After

1. sfdx force:source and force:mdapi commands support DigitalExperienceBundle, DigitalExperience, and DigitalExperienceConfig MD types (without support for force:source:pull, force:source:push and force:source:list)
2. Partial wildcards retrievals for child types are now supported

### DIFF
--- a/METADATA_SUPPORT.md
+++ b/METADATA_SUPPORT.md
@@ -4,7 +4,7 @@ This list compares metadata types found in Salesforce v56 with the [metadata reg
 
 This repository is used by both the Salesforce CLIs and Salesforce's VSCode Extensions.
 
-Currently, there are 465/511 supported metadata types.
+Currently, there are 469/512 supported metadata types.
 For status on any existing gaps, please search or file an issue in the [Salesforce CLI issues only repo](https://github.com/forcedotcom/cli/issues).
 To contribute a new metadata type, please see the [Contributing Metadata Types to the Registry](./contributing/metadata.md)
 
@@ -170,9 +170,9 @@ To contribute a new metadata type, please see the [Contributing Metadata Types t
 |DelegateGroup|✅||
 |DeploymentSettings|✅||
 |DevHubSettings|✅||
-|DigitalExperience|❌|Not supported, but support could be added (but not for tracking)|
-|DigitalExperienceBundle|❌|Not supported, but support could be added (but not for tracking)|
-|DigitalExperienceConfig|❌|Not supported, but support could be added (but not for tracking)|
+|DigitalExperience|⚠️|Supports deploy/retrieve but not source tracking|
+|DigitalExperienceBundle|⚠️|Supports deploy/retrieve but not source tracking|
+|DigitalExperienceConfig|⚠️|Supports deploy/retrieve but not source tracking|
 |DiscoveryAIModel|✅||
 |DiscoveryGoal|✅||
 |DiscoverySettings|✅||
@@ -374,6 +374,7 @@ To contribute a new metadata type, please see the [Contributing Metadata Types t
 |PathAssistantSettings|✅||
 |PaymentGatewayProvider|✅||
 |PaymentsManagementEnabledSettings|✅||
+|PaymentsSettings|✅||
 |PermissionSet|✅||
 |PermissionSetGroup|✅||
 |PermissionSetLicenseDefinition|✅||
@@ -529,6 +530,7 @@ v57 introduces the following new types.  Here's their current level of support
 
 |Metadata Type|Support|Notes|
 |:---|:---|:---|
+|ActionableListDefinition|❌|Not supported, but support could be added|
 
 ## Additional Types
 
--- a/METADATA_SUPPORT.md
+++ b/METADATA_SUPPORT.md
@@ -531,6 +531,9 @@ v57 introduces the following new types.  Here's their current level of support
 |Metadata Type|Support|Notes|
 |:---|:---|:---|
 |ActionableListDefinition|❌|Not supported, but support could be added|
+|ExternalClientApplication|❌|Not supported, but support could be added|
+|GearDescriptor|❌|Not supported, but support could be added|
+|ProductSpecificationTypeDefinition|❌|Not supported, but support could be added|
 
 ## Additional Types
 
--- a/src/collections/componentSet.ts
+++ b/src/collections/componentSet.ts
@@ -523,6 +523,14 @@ export class ComponentSet extends LazyCollection<MetadataComponent> {
         if (parentInWildcard) {
           return true;
         }
+        const partialWildcardKey = this.simpleKey({
+          fullName: `${parent.fullName}.${ComponentSet.WILDCARD}`,
+          type: component.type,
+        });
+        const parentInPartialWildcard = this.components.has(partialWildcardKey);
+        if (parentInPartialWildcard) {
+          return true;
+        }
       }
     }
 
--- a/src/registry/registryAccess.ts
+++ b/src/registry/registryAccess.ts
@@ -129,4 +129,20 @@ export class RegistryAccess {
     }
     return this.aliasTypes;
   }
+
+  /**
+   * Return the parent metadata type from the registry for the given child type
+   *
+   * @param childName - Child metadata type name
+   * @returns Parent metadata type object
+   */
+  public getParentType(childName: string): MetadataType {
+    let parent: MetadataType;
+    const lower = childName.toLowerCase().trim();
+    if (this.registry.childTypes[lower]) {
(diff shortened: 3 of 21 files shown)


METADATA_SUPPORT.md
```diff
@@ -374,6 +374,7 @@ To contribute a new metadata type, please see the [Contributing Metadata Types t
 |PathAssistantSettings|✅||
 |PaymentGatewayProvider|✅||
 |PaymentsManagementEnabledSettings|✅||
+|PaymentsSettings|✅||
```
Kaaviyan, line 377: Running `yarn update-supported-metadata DigitalExperienceBundle DigitalExperience DigitalExperienceConfig` made these changes too. I wasn't sure if I should remove them or just leave them

src/collections/componentSet.ts
```diff
@@ -523,6 +523,14 @@ export class ComponentSet extends LazyCollection<MetadataComponent> {
         if (parentInWildcard) {
           return true;
         }
+        const partialWildcardKey = this.simpleKey({
```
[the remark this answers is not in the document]
Kaaviyan: I added support for this. Incase not supporting this was WAD, do let me know, I'll remove this code from this PR

Kaaviyan:
The CI job fails with `"ERROR: Permission to forcedotcom/source-deploy-retrieve.git denied to testcheckoutkey."` 
Is this expected?

src/registry/registryAccess.ts
```diff
@@ -129,4 +129,20 @@ export class RegistryAccess {
...
+   */
+  public getParentType(childName: string): MetadataType {
+    let parent: MetadataType;
+    const lower = childName.toLowerCase().trim();
+    if (this.registry.childTypes[lower]) {
+      const parentTypeId = this.registry.childTypes[lower];
+      parent = this.registry.types[parentTypeId];
+    }
+    return parent;
+  }
```
mshanemc, line 147: your method returns `MetadataType | undefined`, and that `let` might not ever be assigned but a consumer wouldn't know that if you don't include the  union with `undefined`.

```suggestion
  public getParentType(childName: string): MetadataType | undefined {
    const lower = childName.toLowerCase().trim();
    if (this.registry.childTypes[lower]) {
      const parentTypeId = this.registry.childTypes[lower];
      return this.registry.types[parentTypeId];
    }
  }
```

mshanemc: I wrote this in the browser so it might not work.

src/resolve/metadataResolver.ts
```diff
@@ -297,9 +305,24 @@ export class MetadataResolver {
...
+   * If this file should be considered as a metadata file then return the metadata type
+   */
+  private parseAsMetadata(fsPath: string): string {
+    if (this.tree.isDirectory(fsPath)) {
+      return;
+    }
+    return ['DigitalExperience']
+      .map((type) => this.registry.getTypeByName(type))
+      .find((type) => fsPath.split(sep).includes(type.directoryName))?.name;
+  }
```
shetzel, line 318: `parseAsMetadata` isn't really an accurate name.  This is simply checking the file path to see if it contains the `DigitalExperience` directory.  The more special handling code added to this library the less reliable it becomes so we try to avoid it.

shetzel (changes requested):
I appreciate all the effort with this, including writing tests.  This change has me very worried about how it will affect all other types.  This type you're defining has common elements with existing types that SDR knows how to handle, but differs in significant ways...which is why you had to add another adapter and modify other sensitive areas of the code.  If there is a way to modify things slightly so that it conforms to existing categories of types it would be much preferred.  Can we discuss options before proceeding?

src/resolve/adapters/baseSourceAdapter.ts
```diff
@@ -116,6 +116,10 @@ export abstract class BaseSourceAdapter implements SourceAdapter {
     }
   }
 
+  protected parseMetadataXml(path: SourcePath): MetadataXml {
+    return parseMetadataXml(path);
+  }
```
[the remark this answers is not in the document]
cromwellryan: https://github.com/forcedotcom/source-deploy-retrieve/pull/695/files#diff-5a21a23fb26e4f73dbd10c2582d63cfca25c785fcce9dd57f417ce2283f9b38cR101 I suspect

src/collections/componentSet.ts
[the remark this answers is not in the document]
Kaaviyan: `What would be the difference between asking for site/Xcel_Energy1 of a DEB and site/Xcel_Energy1.* of a DE?`
We won't retrieve -meta.xml of the DEB. You are right that expect this there won't be any other changes, it will only be helpful if there are multiple child types for a parent md type. For DEB there is only one child type for now, so I'm okay to not support this as well

src/resolve/adapters/baseSourceAdapter.ts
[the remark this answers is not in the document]
Kaaviyan: Yes, I needed it to override fullName. We have a full name in the format `type/apiName` where type and api name are individual folders. so I'm using this to construct the proper full name - https://github.com/forcedotcom/source-deploy-retrieve/pull/695/files#diff-5a21a23fb26e4f73dbd10c2582d63cfca25c785fcce9dd57f417ce2283f9b38cR113

src/resolve/metadataResolver.ts
```diff
@@ -215,6 +215,14 @@ export class MetadataResolver {
       resolvedType = this.registry.getTypeBySuffix(extName(fsPath));
     }
 
+    // attempt 4 - try treating the content as metadata
+    if (!resolvedType) {
+      const metadata = this.parseAsMetadata(fsPath);
+      if (metadata) {
+        resolvedType = this.registry.getTypeByName(metadata);
+      }
+    }
```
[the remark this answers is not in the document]
Kaaviyan: This is required for DE. Currently we resolve the type using the following methods and none of it suits DE
1. strict folder - this is where static resource comes in. If I add a strict folder for DE then https://github.com/forcedotcom/source-deploy-retrieve/blob/ec55d3342a3836545c35ec9f01d44b6f81a0c699/src/registry/registryAccess.ts#L95 returns null as DE is a child type and so `registry.types` will not have the type
2. metadata xml - DE is a json file
3. parse as folder - it's not a folder md type
4. by suffix - DE doesn't have a custom suffix
So I had to create a new resolver where we consider a non-xml file (generally identified as a content) as a metadata

src/resolve/metadataResolver.ts
Kaaviyan: As an extension to https://github.com/forcedotcom/source-deploy-retrieve/pull/695#discussion_r962779387 I named it `parseAsMetadata` to have outliers where we don't usually consider these as metadata. In the case of DE, it's a json file where we consider the file as a content on other metadata types. I'm open for suggestions though

src/resolve/adapters/digitalExperienceSourceAdapter.ts
```diff
@@ -0,0 +1,133 @@
...
+ * ├── bars/
+ * |   ├── bars.digitalExperience-meta.xml
+ * ```
+ *
+ * In the above structure the metadata xml file ending with "digitalExperience-meta.xml" belongs to DigitalExperienceBundle MD type.
+ * The "_meta.json" files are child metadata files of DigitalExperienceBundle belonging to DigitalExperience MD type. The rest of the files in the
+ * corresponding folder are the contents to the DigitalExperience metadata. So, incase of DigitalExperience the metadata file is a JSON file
+ * and not an XML file
+ */
+export class DigitalExperienceSourceAdapter extends BundleSourceAdapter {
```
Kaaviyan, line 48: @shetzel just noticed https://github.com/forcedotcom/source-deploy-retrieve/blob/aa79e2642a7aa35b05cb71c9f4cc0b0c21c21123/src/resolve/adapters/index.ts#L7 Should I add this adapter to index.ts?

shetzel: @Kaaviyan - yes, please add it.  Good catch!

shetzel (approved):

PR #748, opened 2022-10-25
Replacements deploy

mshanemc:
### What does this PR do?
string replacements during deploy

define file via glob or filename
define what to replace via exact strings or regex
define new text from env or file
replacements could happen conditional based on one or more envs

provide output in DeployResponse for plugins to use for display

@W-11949818@


src/convert/metadataConverter.ts
```diff
@@ -111,7 +117,10 @@ export class MetadataConverter {
       }
 
       const conversionPipeline = pipeline(
-        new ComponentReader(components),
+        Readable.from(components),
+        !targetFormatIsSource && (process.env.DEBUG_REPLACEMENTS_VIA_CONVERT === 'true' || output.type === 'zip')
```
WillieRuemmele, line 121: want to keep this env var in?

src/convert/replacements.ts
```diff
@@ -0,0 +1,192 @@
...
+/**
+ * Build the replacements property for a sourceComponent
+ */
+export const getReplacements = async (
+  cmp: SourceComponent,
+  replacementConfigs: ReplacementConfig[] = []
+): Promise<SourceComponent['replacements']> => {
+  // all possible filenames for this component
+  const filenames = [cmp.xml, ...cmp.walkContent()].filter(Boolean);
+  // eslint-disable-next-line no-console
```
WillieRuemmele, line 135: not needed?

I'm finding L136-150 really confusing, could be due to GH viewer, but a few comments could go a long way

src/convert/replacements.ts
```diff
@@ -0,0 +1,192 @@
...
+  return fileContentsCache.get(path);
+};
+
+/**
+ * Build the replacements property for a sourceComponent
+ */
+export const getReplacements = async (
+  cmp: SourceComponent,
+  replacementConfigs: ReplacementConfig[] = []
+): Promise<SourceComponent['replacements']> => {
```
WillieRuemmele, line 132: should this be `Optional`? we're conditionally returning

test/nuts/perfResults/x64-linux-2xIntel-Xeon-CPU-E5-2673-v4-2-30GHz/lotsOfClasses.json
```diff
@@ -1,18 +1,18 @@
 [
   {
     "name": "componentSetCreate",
-    "duration": 644.1402209999505
+    "duration": 702.3725049999775
```
WillieRuemmele, line 4: should we add a replacement perf test?

src/convert/metadataConverter.ts
mshanemc: I'm leaning toward yes.  It would also enable a use case I don't want to officially encourage (do replacements to create a mdapi folder that you can mdapi deploy)

mshanemc: but I renamed it to match `sf` style

src/convert/replacements.ts
```diff
@@ -0,0 +1,203 @@
...
+        `Your sfdx-project.json specifies that ${replacement.toReplace.toString()} should be replaced, but it was not found.`
+      );
+    }
+    output = replaced;
+  }
+  return output;
+};
+
+/**
+ * Reads the project, gets replacements, removes an that aren't applicable due ot environment conditionals, and returns an instance of the ReplacementMarkingStream
```
peternhale, line 72: A couple of typos in commend

src/convert/replacements.ts
```diff
@@ -0,0 +1,203 @@
...
+          f,
+          await Promise.all(
+            replacementConfigs
+              // filter out any that don't match the current file
+              .filter((r) => matchesFile(f, r))
+              .map(async (r) => ({
+                // used during replacement stream to limit warnings to explicit filenames, not globs
+                singleFile: Boolean(r.filename),
+                // Config is json which might use the regex.  If so, turn it into an actual regex
+                toReplace: r.stringToReplace ?? new RegExp(r.regexToReplace, 'g'),
```
peternhale, line 150: Line 57 above is looking at toReplace checking if it is a string and if so creating a regex from the string. If toReplace needs to be a regex, why not convert the string into the regex here?

src/convert/replacements.ts
```diff
@@ -0,0 +1,203 @@
...
+    `"${env}" is in sfdx-project.json as a value for "replaceWithEnv" property, but it's not set in your environment.`
+  );
+};
+
+/**
+ * Read the `replacement` property from sfdx-project.json
+ */
+const readReplacementsFromProject = async (): Promise<ReplacementConfig[]> => {
+  const proj = await SfProject.resolve();
+  const projJson = (await proj.resolveProjectConfig()) as { replacements?: ReplacementConfig[] };
```
peternhale, line 201: @mshanemc is your work going to include updating sfdx-project.schema.json in @salesforce/schemas?

src/convert/types.ts
```diff
@@ -101,7 +101,40 @@ export type ConvertResult = {
...
+    ];
+  };
+
+type ReplacementLocation = { filename: string; glob?: never } | { filename?: never; glob: string };
+type ReplacementSource =
+  | { replaceWithEnv: string; replaceWithFile?: never }
+  | { replaceWithEnv?: never; replaceWithFile: string };
+
+type ReplacementTarget =
+  | { stringToReplace: string; regexToReplace?: never }
```
peternhale, line 135: Why carry a string and a regex, when a single property would suffice?

test/convert/replacements.test.ts
```diff
@@ -0,0 +1,234 @@
...
+ */
+import * as path from 'path';
+import { expect } from 'chai';
+import Sinon = require('sinon');
+import { Lifecycle } from '@salesforce/core';
+import { getReplacements, matchesFile, replacementIterations } from '../../src/convert/replacements';
+import { matchingContentFile } from '../mock';
+import * as replacementsForMock from '../../src/convert/replacements';
+
+describe('file matching', () => {
```
peternhale, line 15: Given that regexes are loaded from json, there will be nothing that verifies regexes while modifying the project json, allowing an invalid regex to be saved. At runtime the, the regex will fail parse. How will this error bubble out with enough information to properly identify the source of the error?

peternhale (changes requested):

test/convert/replacements.test.ts
mshanemc:
```
ERROR running force:source:deploy:  Component conversion failed: Invalid regular expression: /BOOM\/: \ at end of pattern
```

How's that?

src/convert/types.ts
mshanemc: exact-match-on-strings is simpler, less error-prone, and require no escaping beyond what json requires for `\`

considering the common scenarios like emails and urls, `user@company.com` or `https://foo.com/api`, it's gonna result in more success for users.

src/convert/replacements.ts
mshanemc: it should.  good call.

src/convert/types.ts
peternhale: @mshanemc Still a bit confused. Can you share some examples in json where expressing a value as a string is simpler than a regex.

peternhale (approved):

WillieRuemmele:
### QA Notes
---
#### Replace with ENV VAR
```
    {
      "glob": "force-app/**/*.cls",
      "stringToReplace": "replaceMeWithEnv",
      "replaceWithEnv": "ENVIRONMENT"
    }
```
✅ : confirmed in the org string was replaced with ENV VAR value
✅ : when `ENVIRONMENT` not set: `ERROR running force:source:deploy:  Component conversion failed: "ENVIRONMENT" is in sfdx-project.json as a value for "replaceWithEnv" property, but it's not set in your environment.`
✅ : in output of `source:push`
```
=== Metadata Replacements

 PROJECT PATH                                        TEXT REPLACED    
 ─────────────────────────────────────────────────── ──────────────── 
 force-app/main/default/classes/GeocodingService.cls replaceMeWithEnv 
```
😐 : in output of `deploy --verbose` identical to `source:push` but not when `--verbose` is excluded
😐 : with env set, but nothing matching in file, no warning nothing was replaced
✅ : new `replacements` in json
✅ : present in `--json` even without specifying `--verbose`
✅ : `source:retrieve` no affect - pulls whats in org
😐 : `"glob": "force-app/**/abc.cls",` with env set, no warning no file was found?
😐 : ` "stringToReplace": "string not in file",` no warning about not finding string?

#### Optionally Replace
```
    {
          "glob": "force-app/**/*.cls",
          "stringToReplace": "replaceThis",
          "replaceWithEnv": "ENVIRONMENT",
          "replaceWhenEnv": [
            {
              "env": "ENVIRONMENT",
              "value": "bar"
            }
          ]
        }
```
✅ : nothing replaced when `ENVIRONEMNT=prod`
✅ : replaced with `bar` when  `ENVIRONEMNT=bar`
⁉️ : invalid regex allowed, `"glob": "[",` should be invalid when parsed as RegEx
#### Replace with File
```
        {
          "glob": "force-app/**/*.cls",
          "stringToReplace": "replaceMeWithFile",
          "replaceWithFile": "replacements.txt"
        }
```
✅ : replaced string with file contents
🏆 : nice error message when file not found `The file "nonexistant.csv" specified in the "replacements" property of sfdx-project.json could not be read.`

#### Replace with RegEx
```
        {
          "glob": "force-app/**/*.cls",
          "regexToReplace": "\\d{4}@W",
          "replaceWithEnv": "ENV"
        }
```
✅ : replaced `// 1234@W` in the Apex Class to `bar` (set at command level)
❓ : had to escape `\` as invalid json so the RegEx became `\\d` to represent `\d` - Should make a note of this when releasing
🏆 : error message indicates invalid RegEx `Invalid regular expression: /[/: Unterminated character class` when `"regexToReplace": "[",`

#### Replace String with File
```
        {
          "glob": "force-app/main/default/classes/**/*",
          "stringToReplace": "find this string",
          "replaceWithFile": "replacements.csv"
        }
```
✅ : replaces all of the matching strings with content of file

#### Replace String with Env Var
```
        {
          "glob": "force-app/main/default/classes/**/*",
          "stringToReplace": "find this string",
          "replaceWithEnv": "ENV"
        }
```
✅ : replaces the string with env var
✅ : replaces all of the strings that match with that env var

WillieRuemmele (approved):

mshanemc:
QA responses:

> invalid regex allowed, "glob": "[", should be invalid when parsed as RegEx

globs are not regex.

> "glob": "force-app/**/abc.cls", with env set, no warning no file was found

yeah, because these are streamed at the component level, there's not some top-level context where the stream is concerned with "has there ever been any file that matched this?"  It's possibly, but not elegantly.

> ❓ : had to escape \ as invalid json so the RegEx became \\d to represent \d - Should make a note of this when releasing

it's going to be in the docs.

PR #780, opened 2022-11-21
fix: add supportsPartialDelete property to registry

shetzel:
### What does this PR do?
Adds a `supportsPartialDelete` property to registry entries to denote that a metadata type should do a deploy when only part of the files that make up the metadata are deleted, rather than a delete.

@W-11385865@
@W-11828599@


WillieRuemmele:
we should add a note in the `adding metadata` section to describe what this new field is, what it denotes, and some examples of metadata types that use it

mshanemc (approved):

PR #817, opened 2023-01-09
fix: pass output dir to SfProject

cristiand391:
### What does this PR do?

Pass the `output` dir from the conversion pipeline to `SfProject.getInstace()` so that it doesn't default to `process.cwd` here:
https://github.com/forcedotcom/sfdx-core/blob/main/src/sfProject.ts#L456-L458

When doing a retrieve like this:

```ts
const mdapiRetrieve = await this.componentSet.retrieve({
  usernameOrConnection: username,
  merge: true,
  output: `/tmp/lalala`
});
```

where:

1. `/tpm/lalala` is where the sfdx project lives (sfdx-project.json file)
2. `process.cwd()` = `/workspace`, no sfdx project.

The retrieve finishes successfully and the conversion pipeline starts, ⬇️ 

SDR is passing the output dir to `ComponentConverter`:
https://github.com/forcedotcom/source-deploy-retrieve/blob/main/src/convert/metadataConverter.ts#L120-L128

but the finalizer for non-decomposed metadata types (CustomLabels) tries to create an instance of `SfProject` that defaults to `process.cwd` which isn't guaranteed to be the root of an sfdx project, thus causing sfdx-core to throw `This directory does not contain a valid Salesforce DX project`

https://github.com/forcedotcom/source-deploy-retrieve/blob/main/src/convert/convertContext.ts#LL204C42-L204C53
https://github.com/forcedotcom/sfdx-core/blob/3349e68b1a0a9cfdf500a2422380dcadae218cd5/src/util/internal.ts#L57

[@W-12360251@](https://gus.my.salesforce.com/a07EE00001J6J7JYAV)

### Functionality Before

<insert gif and/or summary>

### Functionality After

<insert gif and/or summary>

### DIFF
--- a/.github/workflows/onPushToMain.yml
+++ b/.github/workflows/onPushToMain.yml
@@ -2,12 +2,18 @@ name: version, tag and github release
 
 on:
   push:
-    branches: [main]
+    branches:
+      - main
+      - cd/non-decomposed-sf-instance
+    tags-ignore:
+      - '*'
 
 jobs:
   release:
     uses: salesforcecli/github-workflows/.github/workflows/githubRelease.yml@main
     secrets: inherit
+    with:
+      prerelease: ${{ github.ref_name != 'main' }}
   docs:
     uses: salesforcecli/github-workflows/.github/workflows/publishTypedoc.yml@main
     secrets: inherit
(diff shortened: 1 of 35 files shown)


CristiCanizales (approved):
LGTM :)

shetzel (approved):

mshanemc (approved):

PR #839, opened 2023-01-31
fix: respect SF_APPLY_REPLACEMENTS_ON_CONVERT env var

cristiand391:

[skip-validate-pr]

### Functionality Before

<insert gif and/or summary>

### Functionality After

<insert gif and/or summary>

### DIFF
--- a/src/convert/metadataConverter.ts
+++ b/src/convert/metadataConverter.ts
@@ -119,7 +119,7 @@ export class MetadataConverter {
 
       const conversionPipeline = pipeline(
         Readable.from(components),
-        !targetFormatIsSource && (process.env.SF_APPLY_REPLACEMENTS_ON_CONVERT === 'true' || output.type === 'zip')
+        !targetFormatIsSource && output.type === 'zip' && process.env.SF_APPLY_REPLACEMENTS_ON_CONVERT !== 'false'
           ? (await getReplacementMarkingStream(defaultDirectory)) ?? new PassThrough({ objectMode: true })
           : new PassThrough({ objectMode: true }),
         new ComponentConverter(targetFormat, this.registry, mergeSet, defaultDirectory),
--- a/src/convert/metadataConverter.ts
+++ b/src/convert/metadataConverter.ts
@@ -120,7 +120,7 @@ export class MetadataConverter {
       const conversionPipeline = pipeline(
         Readable.from(components),
         !targetFormatIsSource && (process.env.SF_APPLY_REPLACEMENTS_ON_CONVERT === 'true' || output.type === 'zip')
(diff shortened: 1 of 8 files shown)


src/convert/metadataConverter.ts
```diff
@@ -120,7 +120,7 @@ export class MetadataConverter {
       const conversionPipeline = pipeline(
         Readable.from(components),
         !targetFormatIsSource && (process.env.SF_APPLY_REPLACEMENTS_ON_CONVERT === 'true' || output.type === 'zip')
-          ? (await getReplacementMarkingStream()) ?? new PassThrough({ objectMode: true })
+          ? (await getReplacementMarkingStream(defaultDirectory)) ?? new PassThrough({ objectMode: true })
```
cristiand391, line 123: when doing a deploy we need to convert md from source to mdapi format so `targetFormatIsSource === true`, then `getReplacementMarkingStream` is called to return the 2nd transformer for the conversion pipeline.

`getReplacementMarkingStream` calls another func that creates an instance of `SfProject` without passing the path to the sfdx project so it defaults to `process.cwd` which isn't guaranteed to be where the project lives.

src/convert/metadataConverter.ts
```diff
@@ -119,7 +119,7 @@ export class MetadataConverter {
 
       const conversionPipeline = pipeline(
         Readable.from(components),
-        !targetFormatIsSource && (process.env.SF_APPLY_REPLACEMENTS_ON_CONVERT === 'true' || output.type === 'zip')
+        !targetFormatIsSource && output.type === 'zip' && process.env.SF_APPLY_REPLACEMENTS_ON_CONVERT !== 'false'
```
cristiand391, line 122: SDR was still trying to do the string replacements even with the env var set to false:
```
➜  dreamhouse-lwc git:(main) ✗ SF_APPLY_REPLACEMENTS_ON_CONVERT=false sf deploy
metadata -m ApexClass:GeocodingService
Warning: This command is currently in beta. Any aspect of this command can chang
e without advanced notice. Don't use beta commands in your scripts.
*** Deploying v55.0 metadata with SOAP API v57.0 connection ***
Deploy ID: *************
Status: Failed | ████████████████████████████████████████ | 0/0 Components

Component Failures [4]
=====================================================================
| Type  Name             Problem
| ───── ──────────────── ────────────────────────────────────────────
| Error GeocodingService Expecting '}' but was: ';' (2:60)
| Error GeocodingService Variable does not exist: geocodingUrl (2:46)
| Error GeocodingService Missing ';' at '}' (2:58)
| Error GeocodingService Unexpected token '{'. (2:44
```

cristiand391:
nvm, this env var isn't intended to be used to disable string replacements.

PR #847, opened 2023-02-08
fix: only delete non-forceignored sub-directories

shetzel:
### What does this PR do?
When retrieving bundle type metadata that has some pieces of the bundle deleted (or it looks that way because there are more local files than what was retrieved), only delete the pieces that are not forceignored.

@W-12460543@
https://github.com/forcedotcom/cli/issues/1904

### Functionality Before
would delete the entire component and replace with what was retrieved if there was any file within the component locally that was not retrieved, even if forceignored.

### Functionality After
Removes pieces of a component only if they are not forceignored and they were not part of the retrieve.

### DIFF
--- a/src/client/metadataApiRetrieve.ts
+++ b/src/client/metadataApiRetrieve.ts
@@ -367,7 +367,17 @@ export class MetadataApiRetrieve extends MetadataTransfer<
         const localComp = partialDeleteComponents.get(comp.fullName);
         if (localComp.contentPath && tree.isDirectory(comp.content)) {
           const remoteContentList = tree.readDirectory(comp.content);
-          let deleteLocalComp = false;
+
+          const isForceIgnored = (filePath: string): boolean => {
+            const ignored = comp.getForceIgnore().denies(filePath);
+            if (ignored) {
+              this.logger.debug(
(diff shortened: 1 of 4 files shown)


src/client/metadataApiRetrieve.ts
```diff
@@ -376,30 +386,27 @@ export class MetadataApiRetrieve extends MetadataTransfer<
...
                   return;
                 }
+              } else if (isForceIgnored(fileNameFullPath)) {
+                return;
               }
 
               this.logger.debug(
                 `Local component (${comp.fullName}) contains ${fileName} while remote component does not. This file is being removed.`
               );
-              deleteLocalComp = true;
```
WillieRuemmele, line 390: 👍

WillieRuemmele (approved):

WillieRuemmele:
### QA Notes
---

✅ : add a file to a LWC, retrieve it, new file is gone
✅ : add a file to a LWC, forceignore that new file, retrieve the LWC, the file remains
✅ : add a directory, forceignore it, retrieve, directory remains
✅ : add a directory, retrieve, directory is gone

PR #867, opened 2023-02-28
Sm/fix potential undefined org id in sdr event

mshanemc:
### What does this PR do?
when STL subscribes to events, it sets the orgId on the Deploy in the `pre` stage

In SF's use cases, it's possible to do tracking operations async (timeouts, --async) and then `resume` them.  When that happens, the orgId was undefined (because SDR doesn't have strictNullChecks on 😭 ) so the orgId was undefined.

This checks for that case and ensures it is set and used.

### DIFF
--- a/.mocharc.json
+++ b/.mocharc.json
@@ -6,5 +6,5 @@
   "watch-files": ["src", "test"],
   "recursive": true,
   "reporter": "spec",
-  "timeout": 5000
+  "timeout": 10000
 }
--- a/src/client/metadataApiDeploy.ts
+++ b/src/client/metadataApiDeploy.ts
@@ -374,8 +374,8 @@ export class MetadataApiDeploy extends MetadataTransfer<
   // eslint-disable-next-line @typescript-eslint/require-await
   protected async post(result: MetadataApiDeployStatus): Promise<DeployResult> {
     const lifecycle = Lifecycle.getInstance();
+    const connection = await this.getConnection();
     try {
-      const connection = await this.getConnection();
       const apiVersion = connection.getApiVersion();
       // Creates an array of unique metadata types that were deployed, uses Set to avoid duplicates.
       let listOfMetadataTypesDeployed: string[];
@@ -428,7 +428,11 @@ export class MetadataApiDeploy extends MetadataTransfer<
     );
     // only do event hooks if source, (NOT a metadata format) deploy
     if (this.options.components) {
-      await lifecycle.emit('scopedPostDeploy', { deployResult, orgId: this.orgId } as ScopedPostDeploy);
+      // this may not be set if you resume a deploy so that `pre` is skipped.
+      if (!this.orgId) {
+        this.orgId = connection.getAuthInfoFields().orgId;
+      }
+      await lifecycle.emit('scopedPostDeploy', { deployResult, orgId: this.orgId } satisfies ScopedPostDeploy);
     }
     return deployResult;
   }
--- a/src/client/metadataApiDeploy.ts
+++ b/src/client/metadataApiDeploy.ts
@@ -432,7 +432,7 @@ export class MetadataApiDeploy extends MetadataTransfer<
       if (!this.orgId) {
         this.orgId = connection.getAuthInfoFields().orgId;
       }
-      await lifecycle.emit('scopedPostDeploy', { deployResult, orgId: this.orgId } satisfies ScopedPostDeploy);
+      await lifecycle.emit<ScopedPostDeploy>('scopedPostDeploy', { deployResult, orgId: this.orgId });
     }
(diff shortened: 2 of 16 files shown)


src/client/metadataApiDeploy.ts
```diff
@@ -428,7 +428,11 @@ export class MetadataApiDeploy extends MetadataTransfer<
     );
     // only do event hooks if source, (NOT a metadata format) deploy
     if (this.options.components) {
-      await lifecycle.emit('scopedPostDeploy', { deployResult, orgId: this.orgId } as ScopedPostDeploy);
+      // this may not be set if you resume a deploy so that `pre` is skipped.
+      if (!this.orgId) {
+        this.orgId = connection.getAuthInfoFields().orgId;
+      }
+      await lifecycle.emit('scopedPostDeploy', { deployResult, orgId: this.orgId } satisfies ScopedPostDeploy);
```
gbockus-sf, line 435: image
I had to lookup what satisfies does again.  Cool. new and shiny.  For my own understanding why do we need it in this situation? Doesn't the value passed already match the type? I'm not sure why there was an `as` in the first place.

src/client/metadataApiDeploy.ts
```diff
@@ -374,8 +374,8 @@ export class MetadataApiDeploy extends MetadataTransfer<
   // eslint-disable-next-line @typescript-eslint/require-await
   protected async post(result: MetadataApiDeployStatus): Promise<DeployResult> {
     const lifecycle = Lifecycle.getInstance();
+    const connection = await this.getConnection();
```
gbockus-sf, line 377: Why'd you move this outside the try?

.mocharc.json
```diff
@@ -6,5 +6,5 @@
   "watch-files": ["src", "test"],
```
gbockus-sf, line 6: it's a shame we don't have the 'require a WI PR validation check' in this repo.  While annoying at times it's nice to have that consistency of always having an Issue or WI associated with the PR.

gbockus-sf (approved):
On update around type you could make and just some general q's.  
Good to go though I think.

src/client/metadataApiDeploy.ts
shetzel: (unrelated to Gordon's comment) I don't like this entire section.  I don't know what "scopedPostDeploy" is supposed to mean and I think it's too narrow as an event fired from a library.  IMO the library should fire an event called "postDeploy" and the payload contains the data required for listeners to handle it or not.  Otherwise you run into situations where a client wants to subscribe to all deploy events from the library (not just source deploy, but also mdapi deploy) so you need another event for essentially the same thing.

mshanemc: Steve: I generally agree...but...we had an existing postDeploy event. It was typed as `deployResult` (not an object containing a deployResult) so that it's not possible to add properties (like the orgId) without being a breaking change to publicly documented hooks events.

Scoped include the orgId, so that a listener (STL) not only listens for the events, but knows which org they belong to.

src/client/metadataApiDeploy.ts
mshanemc: it's used below, outside the try

gbockus-sf: Is that the right thing to do.  getting a connection seems like a path where an error could be thrown. Was wondering if we shouldn't just get a new connection if we need it below. 
...
but then I took a look at the method and saw we're just eating the exception anyway. :sadness: carry on

src/client/metadataApiDeploy.ts
shetzel: Right, and I think the plugin event is not great either but it was already dealing with toolbelt's event history.  Since we're using core's Lifecycle singleton we need to use a good event naming pattern to avoid collisions, or not use it.  I think you could still name it `sdr:postDeploy`, always fire it and the payload is an object that has all the necessary data for handling.  Not part of this PR, and it would be a major version bump, but I think we should consider it especially if the need arises.

src/client/metadataApiDeploy.ts
mshanemc: logic: if you got to the post stage you definitely had a valid connection (otherwise, how'd you deploy and poll for completion?). Maybe something weird happens between, but probably not.

error swallowing: we don't want to throw on telemetry reporting.

klewis-sfdc (approved):

src/client/metadataApiDeploy.ts
gbockus-sf: If the try/catch is only for telemetry should we put the try on line 392?

mshanemc:
merging--that tracking issue a known change in the sample project

PR #875, opened 2023-03-06
fix(mixedcontentsourceadapter.ts): Error during delete operation of experiencepropertytype metadata bundle

prashanthjos:
### What does this PR do?

This change is to fix the error during the delete operation for experiencepropertytype metadata bundle. As the bundle is all JSON files and there is no XML file in the bundle there will be an error when the delete operation fails in the backend and this is the fix for the same.

This change is to fix the error during the delete operation for experiencepropertytype metadata bundle. As the bundle is all JSON files and there is no XML file in the bundle there will be an error when the delete operation fails in the backend and this is the fix for the same.

#<Insert GitHub Issue> @W-12642608@

### Functionality Before
 Error when the delete operation fails in the backend.

### Functionality After
There will be no error when the delete operation fails in the backend.

### DIFF
--- a/src/registry/metadataRegistry.json
+++ b/src/registry/metadataRegistry.json
@@ -3388,6 +3388,7 @@
       "inFolder": false,
       "strictDirectoryName": true,
       "supportsPartialDelete": true,
+      "metaFileSuffix": "schema.json",
       "strategies": {
         "adapter": "bundle"
       }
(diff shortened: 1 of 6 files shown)


shetzel (changes requested):
The changes look like they should work.  Please add a test for `ExperiencePropertyTypeBundle` to `mixedContentSourceAdapter.test.ts`

shetzel:
merged with PR 885

PR #879, opened 2023-03-10
Sm/strict mode

mshanemc:
> requires https://github.com/salesforcecli/plugin-source/pull/756
> for xNUT purposes, uses https://github.com/salesforcecli/plugin-deploy-retrieve/pull/571.   

### What does this PR do?
resolve some errors that TS strict null checks complain about

breaking:
- make directoryName a required prop on metadataType

other api changes
adds generics on CS.deploy() and metadataApiDeploy to pass to the generic methods on metadataTransfer so that the deploy.id is properly typed (clients won't have to figure out if it's undefined).

[@W-12671663@](https://gus.lightning.force.com/a07EE00001MxCd3YAF)


src/client/diagnosticUtil.ts
```diff
@@ -138,3 +144,11 @@ const parseDefault = (component: SourceComponent, message: DeployMessage): Compo
 
   return diagnostic;
 };
+
+const getProblemFromMessage = (message: string | DeployMessage): string => {
```
mshanemc, line 148: will throw on undefined problem, which the previous version did not, even though it depended it on it being a string it could call `split` on.

src/client/metadataApiRetrieve.ts
```diff
@@ -116,7 +116,10 @@ export class RetrieveResult implements MetadataTransferResult {
     // Add file responses for components that support partial delete (e.g., DigitalExperience)
     // where pieces of the component were deleted in the org, then retrieved.
     while (partialDeleteFileResponses.length) {
-      this.fileResponses.push(partialDeleteFileResponses.pop());
```
mshanemc, line 119: I think this code was fine, but TS wasn't sure that the result of pop would always be defined

shetzel:
All these changes were reviewed and tested in later PRs (e.g. 905) so closing this.

PR #904, opened 2023-03-27
Sm/strict null 5 tests

mshanemc:
### What does this PR do?
Almost all changes are to test classes.
Throws in a lot of asserts about the registry
- the MetadataRegistry is JSON, and the TS types from it are generic across all mdTypes
- therefore, TS can't know that the mdTypes have children (for example).  
- These assertions are pretty low-value, but they do prevent changes to the registry that would break test setup or mocks
Also a lot of asserts about SourceComponent optional properties like xml/content
- also low value, but there's not a clear link between the Registry entry/strategy and the mock files, so this again prevents accidental test breakage and should make it more clear what's wrong with the tests if you do break them (Oh, your mock has no content but the test expects it!)

[@W-12671663@](https://gus.lightning.force.com/a07EE00001MxCd3YAF)

### DIFF
--- a/src/utils/path.ts
+++ b/src/utils/path.ts
@@ -91,7 +91,7 @@ export function parseMetadataXml(fsPath: string): Optional<MetadataXml> {
  * @param directoryName - name of directory to use as a parsing index
  * @returns the FullName
  */
-export function parseNestedFullName(fsPath: string, directoryName: string): Optional<string> {
+export function parseNestedFullName(fsPath: string, directoryName: string): string | undefined {
   const pathSplits = fsPath.split(sep);
   // Exit if the directoryName is not included in the file path.
   if (!pathSplits.includes(directoryName)) {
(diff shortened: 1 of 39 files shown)


src/utils/path.ts
```diff
@@ -91,7 +91,7 @@ export function parseMetadataXml(fsPath: string): Optional<MetadataXml> {
  * @param directoryName - name of directory to use as a parsing index
  * @returns the FullName
  */
-export function parseNestedFullName(fsPath: string, directoryName: string): Optional<string> {
+export function parseNestedFullName(fsPath: string, directoryName: string): string | undefined {
   const pathSplits = fsPath.split(sep);
```
shetzel, line 95: `string | undefined` is the same as `Optional<string>`

mshanemc: yes!  

I mildly prefer the explicit, vanilla TS one to the one that requires an import and makes an external contributor or casual github reader go, "hmm...how does that work?".  It does cost 2 characters more 🤪

PR #906, opened 2023-03-27
Prerelease/set project dir on cs

mshanemc:
### What does this PR do?
devOps center is struggling with cwd being called from Replacements feature which reads the project file via SfProject.
This creates an optional `projectDirectory` prop on ComponentSet which can be used to set the value for SfProject to use instead of cwd.  

[@W-12754654@](https://gus.lightning.force.com/a07EE00001NxDrtYAF)

### DIFF
--- a/.github/workflows/onPushToMain.yml
+++ b/.github/workflows/onPushToMain.yml
@@ -2,12 +2,18 @@ name: version, tag and github release
 
 on:
   push:
-    branches: [main]
+    branches:
+      - main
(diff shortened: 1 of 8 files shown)


WillieRuemmele (approved):

mshanemc:
QA'd by DOCe, closed in favor of https://github.com/forcedotcom/source-deploy-retrieve/pull/908 where we can set a non-prerelease packagejson.version

PR #907, opened 2023-03-27
Prerelease/set-projectDir-on-cs-2

mshanemc:
### What does this PR do?
also handles the retrieve concurrency via the pkgDir on CS


mshanemc:
closed as unneeded per devopsCenter team

PR #913, opened 2023-03-29
ci: metadata support updates as a scheduled job

mshanemc:
updating of  https://github.com/forcedotcom/source-deploy-retrieve/blob/main/METADATA_SUPPORT.md will happen from a daily gha job instead of a commit hook.

[skip-validate-pr]

### DIFF
--- a/.github/workflows/supportedMetadataUpdate.yml
+++ b/.github/workflows/supportedMetadataUpdate.yml
@@ -1,0 +1,22 @@
+name: supported metadata update
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: 3 22 * * *
+
+jobs:
+  registry-check:
+    runs-on: 'ubuntu-latest'
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          token: ${{ secrets.SVC_CLI_BOT_GITHUB_TOKEN }}
+      - uses: salesforcecli/github-workflows/.github/actions/gitConfig@main
+      - uses: actions/setup-node@v3
+        with:
+          node-version: lts/*
+          cache: yarn
+      - run: yarn install  --network-timeout 600000
+      - run: yarn build
+      - run: yarn update-supported-metadata.ts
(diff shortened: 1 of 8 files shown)


.github/workflows/supportedMetadataUpdate.yml
```diff
@@ -0,0 +1,22 @@
...
+        with:
+          token: ${{ secrets.SVC_CLI_BOT_GITHUB_TOKEN }}
+      - uses: salesforcecli/github-workflows/.github/actions/gitConfig@main
+      - uses: actions/setup-node@v3
+        with:
+          node-version: lts/*
+          cache: yarn
+      - run: yarn install  --network-timeout 600000
+      - run: yarn build
+      - run: yarn update-supported-metadata.ts
```
iowillhoit, line 22: Remove the `.ts`

iowillhoit:
Should probably add this to `failureNotifications.yml` otherwise it could fail for months without knowing

iowillhoit:
The `update-supported-metadata.ts` script [does not push the commit](https://github.com/forcedotcom/source-deploy-retrieve/blob/main/scripts/update-registry/update-supported-metadata.ts#L130-L137). I'd consider ripping the `git` logic out of that script and adding it to the github action [like so](https://github.com/salesforcecli/sfdx-cli/blob/main/.github/workflows/update-docker-node-version.yml#L25-L31).

Question: do we want this to be a `fix:` commit so that it releases a new `patch`?

mshanemc:
> The `update-supported-metadata.ts` script [does not push the commit](https://github.com/forcedotcom/source-deploy-retrieve/blob/main/scripts/update-registry/update-supported-metadata.ts#L130-L137). I'd consider ripping the `git` logic out of that script and adding it to the github action [like so](https://github.com/salesforcecli/sfdx-cli/blob/main/.github/workflows/update-docker-node-version.yml#L25-L31).
good idea

> Question: do we want this to be a `fix:` commit so that it releases a new `patch`?
no, this is unrelated to releases and the metadataSupport file doesn't ship in a release.  It's docs-only.

iowillhoit (approved):

PR #941, opened 2023-04-18
fix: retry on net errors when retrieving MD info

cristiand391:
### What does this PR do?

Make SDR retry on network errors like ETIMEDOUT when querying info about metadata types.

@W-13056201@

### Functionality Before

`force source manifest create` would exit successfully with an incomplete manifest because SDR ignores errors when getting MD info.

### Functionality After
`SDR` retries on network errors, if polling client times out then throw error (sfdx should catch the error and fail).

### DIFF
--- a/src/resolve/connectionResolver.ts
+++ b/src/resolve/connectionResolver.ts
@@ -112,6 +112,10 @@ export class ConnectionResolver {
     try {
       members = await pollingClient.subscribe();
     } catch (error) {
+      // throw error if PollingClient timed out.
+      if (error instanceof NotRetryableError) {
+        throw NotRetryableError;
+      }
       this.logger.debug((error as Error).message);
       members = [];
     }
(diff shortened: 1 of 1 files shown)


src/resolve/connectionResolver.ts
```diff
@@ -98,9 +98,24 @@ export class ConnectionResolver {
...
+
+    const pollingClient = await PollingClient.create(pollingOptions);
+
     try {
-      members = ensureArray((await this.connection.metadata.list(query)) as FileProperties[]);
+      members = await pollingClient.subscribe();
     } catch (error) {
+      // throw error if PollingClient timed out.
+      if (error instanceof NotRetryableError) {
+        throw NotRetryableError;
```
cristiand391, line 117: this catch block ignores error on purpose (see lines below) so we need to check if the error comes from the polling client.

shetzel (approved):

PR #975, opened 2023-05-18
Sh/archiver to jszip swap

shetzel:
### What does this PR do?
Replaces the archiver library with jszip.

@W-13196750@

QA Notes:
Be sure to test using both NodeJS v18.16.0 and v18.15.0, but primarily 18.16.0.
1. conversion of static resources
2. deploys, both source and metadata format with static resources
3. retrieves
4. package version create with static resources

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,7 +70,7 @@
     "eslint-plugin-jsdoc": "^39.9.1",
     "eslint-plugin-sf-plugin": "^1.15.1",
     "husky": "^7.0.4",
-    "jsforce": "^2.0.0-beta.21",
+    "jsforce": "^2.0.0-beta.23",
     "mocha": "^9.2.2",
     "mocha-junit-reporter": "^1.23.3",
     "nyc": "^15.1.0",
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
     "got": "^11.8.6",
     "graceful-fs": "^4.2.11",
     "ignore": "^5.2.4",
+    "jszip": "^3.10.1",
     "mime": "2.6.0",
     "minimatch": "^5.1.6",
     "proxy-agent": "^5.0.0",
--- a/package.json
+++ b/package.json
@@ -46,7 +46,6 @@
     "@salesforce/dev-scripts": "^4.3.0",
     "@salesforce/prettier-config": "^0.0.2",
     "@salesforce/ts-sinon": "^1.4.6",
-    "@types/archiver": "^5.3.1",
     "@types/deep-equal-in-any-order": "^1.0.1",
     "@types/fast-levenshtein": "^0.0.2",
     "@types/graceful-fs": "^4.1.6",
--- a/package.json
+++ b/package.json
@@ -130,7 +130,9 @@
       "files": [
         "src/**/*.ts",
         "**/tsconfig.json",
-        "messages/**"
+        "messages/**",
+        "src/registry/metadataRegistry.json",
+        "src/registry/stdValueSetRegistry.json"
       ],
       "output": [
         "lib/**",
@@ -185,7 +187,9 @@
         "**/tsconfig.json",
         ".mocha*",
         "!*.nut.ts",
-        ".nycrc"
+        ".nycrc",
+        "src/registry/metadataRegistry.json",
+        "src/registry/stdValueSetRegistry.json"
       ],
       "output": []
     }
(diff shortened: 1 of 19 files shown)


package.json
```diff
@@ -28,12 +28,12 @@
     "@salesforce/core": "^3.36.2",
     "@salesforce/kit": "^1.9.2",
     "@salesforce/ts-types": "^1.7.2",
-    "archiver": "^5.3.1",
```
mshanemc, line 31: be sure to take the types out of devDeps

src/client/metadataApiDeploy.ts
```diff
@@ -437,20 +436,38 @@ export class MetadataApiDeploy extends MetadataTransfer<
...
-      // the typings on it are misleading and unintended.  More info https://github.com/archiverjs/node-archiver/issues/476
-      // If you await it, bad things happen, like the convert process exiting silently.  https://github.com/forcedotcom/cli/issues/1791
-      // leave the void as it is
-      void zip.finalize();
-      return stream2buffer(zip);
+
+      const zip = JSZip();
+
+      const zipDirRecursive = (dir: string): void => {
+        const list = fs.readdirSync(dir);
```
mshanemc, line 448: `readdir` has an option for [`withFileTypes`](https://nodejs.org/api/fs.html#fspromisesreaddirpath-options) that will give you the isDirectory value without having to do _another_ fs operation inside the loop.

src/convert/streams.ts
```diff
@@ -183,19 +183,17 @@ export class StandardWriter extends ComponentWriter {
...
-  private zip: Archiver = createArchive('zip', { zlib: { level: 3 } });
-  private buffers: Buffer[] = [];
+  private zip = JSZip();
+  private zipBuffer?: Buffer;
 
   public constructor(rootDestination?: SourcePath) {
     super(rootDestination);
-    void pipeline(this.zip, this.getOutputStream());
+    const destination = rootDestination ? `for: ${rootDestination}` : 'in memory';
+    this.logger.debug(`generating zip ${destination}`);
```
mshanemc, line 192: if the logger is only used in ZipWriter, why not create it on the lowest level so all the other writers don't suffer the perf impact of unused loggers.

then, if someone else needs it, they could hoist it higher in the hierarchy

src/convert/streams.ts
```diff
@@ -210,10 +208,7 @@ export class ZipWriter extends ComponentWriter {
           }
           // everything else can be zipped immediately to reduce the number of open files (windows has a low limit!) and help perf
           const streamAsBuffer = await stream2buffer(writeInfo.source);
-          return streamAsBuffer.length
-            ? this.addToZip(streamAsBuffer, writeInfo.output)
-            : // these will be zero-length files, which archiver supports via stream but not buffer
-              this.addToZip(writeInfo.source, writeInfo.output);
+          return this.addToZip(streamAsBuffer, writeInfo.output);
```
mshanemc, line 211: cool!

mshanemc (changes requested):

src/convert/streams.ts
shetzel: It's also used in `StandardWriter`

mshanemc (approved):
commits seem fine, no obvious perf impacts (looking especially at the `sourceToZip` across both OS).  

https://forcedotcom.github.io/source-deploy-retrieve/perf-Linux/
https://forcedotcom.github.io/source-deploy-retrieve/perf-Windows/

mshanemc:
QA notes:
✅ deploy a static resource (single file) to org
✅ deploy a static resource (single file, but empty) to org

✅ deploy a zip containing one empty text file
✅  deploy an empty static resource folder as a zip (same error as current SDR/PDR combo)
| Error testBasic The specified Static Resource is not a valid zip file 

📓  Scenario: static resource has 1 zip file in it.  That file is forceIgnored.  [we got an issue about this scenario once](https://github.com/forcedotcom/source-deploy-retrieve/pull/663/files), and I saw the code change related to it.

current SDR/PDR throws this error **before sending anything to the API**.  
```Error (1): Metadata API request failed: force-app/main/default/staticresources/testBasic.resource-meta.xml: Expected source files for type 'StaticResource'```

👎🏻  this PR sends the bad zip file to the API, get a deployment failure, and shows this error.  I slightly prefer the old way (fail faster, especially if it's a large deployment)
```Error (1): Metadata API request failed: force-app/main/default/staticresources/testBasic.resource-meta.xml: Expected source files for type 'StaticResource'```

Still net improvement over archiver and happy to ship it.

PR #987, opened 2023-05-23
fix: update registry for ExpressionSetVersion

randi274:
### What does this PR do?
Follow-up PR to #758 we had received an investigation that ExpressionSetVersion could not be retrieved separately (it was only coming through as child of the parent ExpressionSetDefinitionVersion. I removed the manual changes from that PR, ran the `update-registry` script, and both metadata types are now being retrieved as expected.

@W-13207993@

### Functionality Before
ExpressionSetVersion (ExpressionSetDefinitionVersion) was not being retrieved.
https://github.com/forcedotcom/source-deploy-retrieve/assets/9795193/7aead06e-94ff-4095-8fb2-3bb6cee46323

### Functionality After
This metadata is now retrieved through VS Code's Org Browser & Manifest Builder as expected.
https://github.com/forcedotcom/source-deploy-retrieve/assets/9795193/0b937b3a-618f-463b-b183-5eb3bbcfb1d0

### DIFF
--- a/src/registry/metadataRegistry.json
+++ b/src/registry/metadataRegistry.json
@@ -2513,6 +2513,22 @@
         }
       }
     },
+    "expressionsetdefinition": {
+      "id": "expressionsetdefinition",
+      "name": "ExpressionSetDefinition",
+      "suffix": "expressionSetDefinition",
+      "directoryName": "expressionSetDefinition",
+      "inFolder": false,
+      "strictDirectoryName": false
+    },
+    "expressionsetdefinitionversion": {
+      "id": "expressionsetdefinitionversion",
+      "name": "ExpressionSetDefinitionVersion",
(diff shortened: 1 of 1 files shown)


randi274:
After discussing with @mshanemc, we decided it would be a better approach to return to the parent/child relationship of the initial PR, and model it after `bot` and `botVersion`. We weren't able to get it quite right during a working session, so the BRE Leo team will follow up in a subsequent WI after confirming with the Metadata API team about best practices on how to structure the Definition/Version. There will still be some temporary discomfort around using ExpressionSetVersion until that's fixed, and we recommend customers continue using the internal [package manager](https://help.salesforce.com/s/articleView?id=sf.expression_set_migration_considerations.htm&type=5) rather than the CLI/VS Code to deploy versions.

PR #990, opened 2023-05-25
fix!: use more jsforce typings

mshanemc:
### What does this PR do?
SDR has TS types for all the mdapi responses.  One of them was wrong: 
https://github.com/salesforcecli/plugin-deploy-retrieve/pull/633
forcedotcom/cli#2149 @W-13481008@

while we're making a major version, I took several jsforce2 types and used them for SDR.  There's a few differences/corrections, so wanted to get this in the major version.

jsforce2 types are generated from the wsdl, so they tend to be more accurate except for array types which it can't tell are optional or not.  If it weren't for that problem, I'd use jsforce's types even more extensively.


mshanemc:
nope!

PR #1031, opened 2023-07-06
Sh/add aqlmuf interval standard value set

shetzel:
### What does this PR do?
change `AQLMUFInterval` to `AQLMFrequencyInterval` for StandardValueSet

https://github.com/forcedotcom/cli/issues/2163


pgumiela:
@shetzel,
Hi Steve,
Don't we actually need both of them? AQLMUFInterval and AQLMFrequencyInterval? Because there are two different standard picklists in ASM:
- AdQuoteLine.MaximumFrequencyInterval (AQLMFrequencyInterval)
- AdQuoteLine.MaximumUserFrequencyInterval (AQLMUFInterval)

And in this PR we only cover the first one - AQLMFrequencyInterval.

PR #1062, opened 2023-08-01
fix: throw when expected file is ENOENT

mshanemc:
### What does this PR do?
SDR would exit streams early on ENOENT without throwing an error
this checks file existence before doing a readStream from the file

[this would be more efficient, on the happy path, to catch this error but it seems to buried inside nested streams and this is a simple fix.  It *would* error much earlier in the convert process if the error is going to happen.  Perf benchmarks suggest it has a negative impact]

https://github.com/forcedotcom/cli/issues/2011
@W-12769219@

### DIFF
--- a/src/resolve/treeContainers.ts
+++ b/src/resolve/treeContainers.ts
@@ -113,6 +113,9 @@ export class NodeFSTreeContainer extends TreeContainer {
   }
 
   public stream(fsPath: SourcePath): Readable {
+    if (!this.exists(fsPath)) {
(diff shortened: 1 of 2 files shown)


shetzel (approved):

shetzel:
With the changes I now see:
`~/dev/salesforcecli/plugin-deploy-retrieve/bin/dev project deploy start
Error (1): Component conversion failed: File not found: ...dreamhouse-lwc/force-app/main/default/classes/FileUtilities.cls-meta.xml`

PR #1070, opened 2023-08-08
fix: match using decoded component keys in ComponentSet.has()

shetzel:
### What does this PR do?
Allows `ComponentSet.has()` to match an entry with an encoded component key by first decoding it.  For example, if a component set has a layout with a fullName of `AssignmentHistory__c-v1.34 Assignment History Layout` and we ask that component set if it has an entry with a fullName of `AssignmentHistory__c-v1%2E34 Assignment History Layout`, it will now return `true`.

@W-11658886@
https://github.com/forcedotcom/cli/issues/1683

### Functionality Before
Remote changed layouts with encoded API names would not match existing local layouts, and would be written to the default directory as a new file.

### Functionality After
Remote changed layouts with encoded API names now can match existing local layouts, resulting in the layout content being updated in the correct file.


src/collections/componentSet.ts
```diff
@@ -547,6 +547,13 @@ export class ComponentSet extends LazyCollection<MetadataComponent> {
       return true;
     }
 
+    // This will decode the key of the component before comparing, which can solve some edge cases
+    // in component fullNames such as Layouts. See: https://github.com/forcedotcom/cli/issues/1683
+    const isDirectlyInSetDecoded = this.components.has(decodeURI(simpleKey(component)));
+    if (isDirectlyInSetDecoded) {
+      return true;
+    }
```
mshanemc, line 555: were you originally doing something else with `isDirectlyInSetDecoded` and this is all that's left?
```suggestion
    if (this.components.has(decodeURI(simpleKey(component)))) {
      return true;
    }
```

mshanemc (approved):
no objections, 1 suggestion

src/collections/componentSet.ts
shetzel: I was (lazily) following the pattern from above.  I saw this after I put the PR up and forgot to go back and change it.  I'll change it in both spots.

mshanemc (approved):

PR #1080, opened 2023-08-15
feat!: component set components are now DecodableMaps

shetzel:
### What does this PR do?
Creates a new DecodeableMap class that extends the standard Map.  A DecodableMap is the same as a Map except that the key must be a string, and whenever passing a key it will match on the exact text and the decoded text.  Since `ComponentSet.components` Maps use component fullNames as keys, this allows encoded fullNames to match decoded fullNames, and vice versa.

 @W-11658886@

### DIFF
--- a/src/collections/componentSet.ts
+++ b/src/collections/componentSet.ts
@@ -44,6 +44,78 @@ export type RetrieveSetOptions = Omit<MetadataApiRetrieveOptions, 'components'>;
 const KEY_DELIMITER = '#';
 
 /**
+ * This is an extension of the Map class that treats keys as the same by matching first normally,
+ * then decoded. Decoding the key before comparing can solve some edge cases in component fullNames
+ * such as Layouts. See: https://github.com/forcedotcom/cli/issues/1683
+ *
+ * Examples:
+ *
+ * Given a map with entries:
+ * ```javascript
+ * 'layout#Layout__Broker__c-v1%2E1 Broker Layout' : {...}
+ * 'layout#Layout__Broker__c-v9.2 Broker Layout' : {...}
+ * ```
+ *
+ * `decodeableMap.has('layout#Layout__Broker__c-v1.1 Broker Layout')` --> returns `true`
+ * `decodeableMap.has('layout#Layout__Broker__c-v9%2E2 Broker Layout')` --> returns `true`
+ */
+class DecodeableMap<K extends string, V> extends Map<string, V> {
+  /**
+   * boolean indicating whether an element with the specified key (matching decoded) exists or not.
+   */
+  public has(key: K): boolean {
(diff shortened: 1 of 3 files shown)


src/collections/componentSet.ts
```diff
@@ -43,6 +43,78 @@ export type RetrieveSetOptions = Omit<MetadataApiRetrieveOptions, 'components'>;
...
+ * Given a map with entries:
+ * ```javascript
+ * 'layout#Layout__Broker__c-v1%2E1 Broker Layout' : {...}
+ * 'layout#Layout__Broker__c-v9.2 Broker Layout' : {...}
+ * ```
+ *
+ * `decodeableMap.has('layout#Layout__Broker__c-v1.1 Broker Layout')` --> returns `true`
+ * `decodeableMap.has('layout#Layout__Broker__c-v9%2E2 Broker Layout')` --> returns `true`
+ */
+class DecodeableMap<K extends string, V> extends Map<string, V> {
```
mshanemc, line 62: suggestion: put this in a separate file and import into CS.  CS is a big file already.

then it's also be easier to throw some UT at without the complexity of CS

src/collections/componentSet.ts
```diff
@@ -43,6 +43,78 @@ export type RetrieveSetOptions = Omit<MetadataApiRetrieveOptions, 'components'>;
...
+    return existingKey ? super.get(existingKey) : undefined;
+  }
+
+  // Returns the key as it is in the map, matching on decoded keys.
+  private getExistingKey(key: string): string | undefined {
+    for (const compKey of this.keys()) {
+      if (decodeURIComponent(compKey) === decodeURIComponent(key)) {
+        return compKey;
+      }
+    }
```
mshanemc, line 114:
```suggestion
    return this.keys().find((compKey) => decodeURIComponent(compKey) === decodeURIComponent(key))
```

mshanemc (approved):

src/collections/componentSet.ts
shetzel: `this.keys()` returns an `IterableIterator<string>` so `find()` isn't available.

src/collections/componentSet.ts
shetzel: I like it.  Moving.

src/collections/componentSet.ts
mshanemc: > this.keys() returns an IterableIterator<string> so find() isn't available.

you can call Array.from() on those to get a real array.

`return Array.from(this.keys()).find((compKey) => decodeURIComponent(compKey) === decodeURIComponent(key))

mshanemc:
QA is here: https://github.com/forcedotcom/source-tracking/pull/458

mshanemc:
Approved for release.

PR #1093, opened 2023-09-01
Cd/forceignore bug

cristiand391:
### What does this PR do?

Updates the metadata resolver to skip the first forceignore check when resolving components recursively if the dir is a folder.

in the linked GH issue below, when building the component set from the API response SDR is trying checking if [`unpackaged`](https://developer.salesforce.com/docs/atlas.en-us.api_meta.meta/api_meta/meta_retrieve_request.htm) should be ignored here:
https://github.com/forcedotcom/source-deploy-retrieve/blob/f288319e55378d1a7caca5bcd762b77b2c59f30a/src/resolve/metadataResolver.ts#L74

`forceignore.denies` will pass a relative path to `node-ignore` that doesn't have a trailing slash so it is treated as a file when matching the forceignore rules against it:
https://github.com/forcedotcom/source-deploy-retrieve/blob/f288319e55378d1a7caca5bcd762b77b2c59f30a/src/resolve/forceIgnore.ts#L66 

https://github.com/forcedotcom/cli/issues/2404
@W-13948748@

### DIFF
--- a/src/resolve/metadataResolver.ts
+++ b/src/resolve/metadataResolver.ts
@@ -71,7 +71,11 @@ export class MetadataResolver {
     const components: SourceComponent[] = [];
     const ignore = new Set();
 
-    if (this.tree instanceof NodeFSTreeContainer && this.forceIgnore?.denies(dir)) {
+    // don't apply forceignore rules against dirs
+    // `forceignore.denies` will pass a relative path to node-ignore, e.g.
+    // `path/to/force-app` -> `force-app`, note that there's no trailing slash
+    // so node-ignore will treat it as a file.
+    if (!this.tree.isDirectory(dir) && this.forceIgnore?.denies(dir)) {
       return components;
     }
 
--- a/src/resolve/metadataResolver.ts
+++ b/src/resolve/metadataResolver.ts
@@ -71,7 +71,7 @@ export class MetadataResolver {
     const components: SourceComponent[] = [];
     const ignore = new Set();
 
-    if (this.forceIgnore?.denies(dir)) {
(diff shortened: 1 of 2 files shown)


src/resolve/metadataResolver.ts
```diff
@@ -71,7 +71,11 @@ export class MetadataResolver {
     const components: SourceComponent[] = [];
     const ignore = new Set();
 
-    if (this.forceIgnore?.denies(dir)) {
+    // don't apply forceignore rules against dirs
+    // `forceignore.denies` will pass a relative path to node-ignore, e.g.
+    // `path/to/force-app` -> `force-app`, note that there's no trailing slash
+    // so node-ignore will treat it as a file.
+    if (!this.tree.isDirectory(dir) && this.forceIgnore?.denies(dir)) {
```
cristiand391, line 78: @mshanemc it turns out the bug also affects a real FS tree:
Screenshot 2023-09-01 at 18 51 35

`path.relative` doesn't return a trailing slash so `node-ignore` will treat as a file, called here:
https://github.com/forcedotcom/source-deploy-retrieve/blob/f288319e55378d1a7caca5bcd762b77b2c59f30a/src/resolve/forceIgnore.ts#L66C1-L66C1

another option I showed you this morning was to add the trailing slash before passing `dir` to `denies`, but then have to remove it again or it will not be able to file the folder in the tree:
https://github.com/forcedotcom/source-deploy-retrieve/blob/f288319e55378d1a7caca5bcd762b77b2c59f30a/src/resolve/metadataResolver.ts#L78

that also worked, but not sure if it's the right solution:
https://stackoverflow.com/questions/980255/should-a-directory-path-variable-end-with-a-trailing-slash

mshanemc (approved):

PR #1159, opened 2023-11-01
fix(lint): current linter rules

mshanemc:
### What does this PR do?
cleans up some things the linter didn't like.  Most of them are related to using truthiness on things that can be undefined in stuff chained by `||`.  A few were actual problems...the others are a bit much.  The logic was fine, but harder for the linter to know, with non-booleans, whether the `||` was being used a logical OR vs. fallback if falsy.

kinda fallout from [@W-14373783@](https://gus.lightning.force.com/a07EE00001dDMpBYAW)
https://github.com/forcedotcom/eslint-config-salesforce-typescript/actions/runs/6720828841/job/18265239393

### DIFF
--- a/src/convert/replacements.ts
+++ b/src/convert/replacements.ts
@@ -195,7 +195,8 @@ export const getReplacements = async (
 export const matchesFile = (f: string, r: ReplacementConfig): boolean =>
   // filenames will be absolute.  We don't have convenient access to the pkgDirs,
   // so we need to be more open than an exact match
-  Boolean((r.filename && posixifyPaths(f).endsWith(r.filename)) || (r.glob && minimatch(f, `**/${r.glob}`)));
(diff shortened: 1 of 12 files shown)


src/convert/replacements.ts
```diff
@@ -195,7 +195,8 @@ export const getReplacements = async (
 export const matchesFile = (f: string, r: ReplacementConfig): boolean =>
   // filenames will be absolute.  We don't have convenient access to the pkgDirs,
   // so we need to be more open than an exact match
-  Boolean((r.filename && posixifyPaths(f).endsWith(r.filename)) || (r.glob && minimatch(f, `**/${r.glob}`)));
+  (typeof r.filename === 'string' && posixifyPaths(f).endsWith(r.filename)) ||
```
mshanemc, line 198: undefined && something => undefined

So it's possible this could return undefined if they're both undefined.

WillieRuemmele (approved):

PR #1191, opened 2023-12-13
fix: change mdcoverage url for getCurrentApiVersion

shetzel:
### What does this PR do?
Changes the URL to get an API version from https://mdcoverage.secure.force.com/services/apexrest/report to https://dx-extended-coverage.my.salesforce-sites.com/services/data
If neither `sourceApiVersion` nor `apiVersion` were set on a `ComponentSet` before calling `getObject()`, will then look for and use (in order):
1. sourceApiVersion set in sfdx-project.json
2. apiVersion from ConfigAggregator (config files and Env Vars)
3. http call to /services/data endpoint for highest apiVersion
4. hardcoded value of "58.0" as a last resort

@W-14621701@

### DIFF
--- a/src/collections/componentSet.ts
+++ b/src/collections/componentSet.ts
@@ -369,7 +369,17 @@ export class ComponentSet extends LazyCollection<MetadataComponent> {
    * @returns Object representation of a package manifest
    */
   public async getObject(destructiveType?: DestructiveChangesType): Promise<PackageManifestObject> {
-    const version = this.sourceApiVersion ?? this.apiVersion ?? `${await getCurrentApiVersion()}.0`;
+    let version: string;
+    try {
+      version =
+        this.sourceApiVersion ??
+        this.apiVersion ??
+        process.env.SF_ORG_API_VERSION ??
+        `${await getCurrentApiVersion()}.0`;
+    } catch (err: unknown) {
+      const cause = err instanceof Error ? err : isString(err) ? Error(err) : undefined;
+      throw messages.createError('missingApiVersion', undefined, undefined, 1, cause);
+    }
 
     // If this ComponentSet has components marked for delete, we need to
     // only include those components in a destructiveChanges.xml and
--- a/src/collections/componentSet.ts
+++ b/src/collections/componentSet.ts
@@ -6,7 +6,16 @@
  */
 /* eslint  @typescript-eslint/unified-signatures:0 */
 import { XMLBuilder } from 'fast-xml-parser';
-import { AuthInfo, Connection, Logger, Messages, SfError } from '@salesforce/core';
+import {
+  AuthInfo,
+  ConfigAggregator,
+  Connection,
+  Logger,
+  Messages,
+  OrgConfigProperties,
+  SfError,
+  SfProject,
+} from '@salesforce/core';
 import { isString } from '@salesforce/ts-types';
 import {
   MetadataApiDeploy,
@@ -369,17 +378,7 @@ export class ComponentSet extends LazyCollection<MetadataComponent> {
    * @returns Object representation of a package manifest
    */
   public async getObject(destructiveType?: DestructiveChangesType): Promise<PackageManifestObject> {
-    let version: string;
-    try {
-      version =
-        this.sourceApiVersion ??
-        this.apiVersion ??
-        process.env.SF_ORG_API_VERSION ??
-        `${await getCurrentApiVersion()}.0`;
-    } catch (err: unknown) {
-      const cause = err instanceof Error ? err : isString(err) ? Error(err) : undefined;
-      throw messages.createError('missingApiVersion', undefined, undefined, 1, cause);
-    }
+    const version = await this.getApiVersion();
 
     // If this ComponentSet has components marked for delete, we need to
     // only include those components in a destructiveChanges.xml and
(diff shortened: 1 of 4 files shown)


src/collections/componentSet.ts
```diff
@@ -369,7 +369,17 @@ export class ComponentSet extends LazyCollection<MetadataComponent> {
    * @returns Object representation of a package manifest
    */
   public async getObject(destructiveType?: DestructiveChangesType): Promise<PackageManifestObject> {
-    const version = this.sourceApiVersion ?? this.apiVersion ?? `${await getCurrentApiVersion()}.0`;
+    let version: string;
+    try {
+      version =
+        this.sourceApiVersion ??
+        this.apiVersion ??
+        process.env.SF_ORG_API_VERSION ??
```
mshanemc, line 377: would you also consider using ConfigAggregator here?  (it would check local/global config AND the env)?  Like that would be to check local FS before making a callout?

mshanemc (approved):

src/collections/componentSet.ts
```diff
@@ -662,6 +671,52 @@ export class ComponentSet extends LazyCollection<MetadataComponent> {
...
+        const projectConfig = await project.resolveProjectConfig();
+        version = projectConfig?.sourceApiVersion as string;
+      } catch (e) {
+        // If there's any problem just move on to ConfigAggregator
+      }
+    }
+
+    if (!version) {
+      try {
+        version = ConfigAggregator.getValue(OrgConfigProperties.ORG_API_VERSION).value as string;
```
mshanemc, line 704: I think you can say `ConfigAggregator.getValue<string>`

mshanemc (approved):
comment is nit.

I'd suggest recommend an FYI to VSCode team about this change--they *should* be doing apiVersion stuff already.  I just know they've had some issues in the past around these Singleton and `cwd` patterns in the past when they have long-running processes.  Give Pete's folks something else to test.

src/collections/componentSet.ts
shetzel: apparently that's `getPropertyValue()`

test/collections/componentSet.test.ts
```diff
@@ -725,6 +737,87 @@ describe('ComponentSet', () => {
...
+            {
+              name: registry.types.customobjecttranslation.name,
+              members: ['a'],
+            },
+            {
+              name: registry.types.staticresource.name,
+              members: ['b', 'c'],
+            },
+          ],
+          version: '58.0',
```
diyer, line 817: why version 58.0 as default?

shetzel: I just chose a recent one.  It's better than throwing an error.  We really should never get to this point.  ComponentSets should have apiVersion and sourceApiVersion set by the consumer.

src/registry/coverage.ts
```diff
@@ -23,7 +23,7 @@ const getProxiedOptions = (url: string): OptionsOfTextResponseBody => ({
 
 export const getCurrentApiVersion = async (): Promise<number> =>
   (
-    await got(getProxiedOptions('https://mdcoverage.secure.force.com/services/apexrest/report')).json<{
+    await got(getProxiedOptions('https://dx-extended-coverage.my.salesforce-sites.com/services/apexrest/report')).json<{
```
peternhale, line 26: @shetzel I don't think this going to work. I'll DM you with the reasoning.

peternhale (changes requested):

test/collections/componentSet.test.ts
diyer: At some point version 58.0 will be too old and things may not function as expected on this version. In that case does it make sense to throw an error vs using a default value that is old?

shetzel: The code will warn with the `missingApiVersion` message should it ever get to this point and use v58.0, so if an end user sees this message they can do things to get better results should it fail.  If we throw, they will definitely need to do something to get it to work.  Both Shane and I felt that doing what we could to not fail is better.

In practice, the checks for an API version should never get past the `sourceApiVersion` and `apiVersion` set on the `ComponentSet`.  If users of SDR are creating ComponentSets without setting a `sourceApiVersion` and/or `apiVersion` then they shouldn't be creating a manifest from it.

peternhale (approved):

src/registry/coverage.ts
```diff
@@ -21,12 +21,17 @@ const getProxiedOptions = (url: string): OptionsOfTextResponseBody => ({
...
-    }>()
-  ).versions.selected;
+type ApiVersion = {
+  label: string;
+  url: string;
+  version: string;
+};
+
+export const getCurrentApiVersion = async (): Promise<number> => {
+  const apiVersionsUrl = 'https://dx-extended-coverage.my.salesforce-sites.com/services/data';
```
mshanemc, line 31: If we're taking that approach, maybe `https://org62.my.salesforce-sites.com/services/data` would be more reliable/scalable than this team's endpoint.

or use `Promise.any` of several known URLs

mshanemc (approved):

mshanemc:
QA (using the node REPL)

```
> const csb = require('./lib/src/collections/componentSetBuilder.js').ComponentSetBuilder
undefined
> const result = await csb.build({sourcepath: ['./test/nuts/local/replacements/testProj/force-app']})
undefined
> const pkg = await result.getPackageXml()
undefined
> pkg
```
✅ has `  '    <version>59.0</version>\n' +`
[turns off internet]
```
> const pkg2 = await result.getPackageXml()
undefined
> pkg2
```
✅ defaulted to v58   '    <version>58.0</version>\n' +

PR #1192, opened 2023-12-14
fix: implicit dependency

mshanemc:
### What does this PR do?
missing dep for ts-retry-promise

[skip-validate-pr]

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,8 @@
     "jszip": "^3.10.1",
     "mime": "2.6.0",
     "minimatch": "^5.1.6",
-    "proxy-agent": "^6.3.1"
+    "proxy-agent": "^6.3.1",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -36,7 +36,8 @@
     "jszip": "^3.10.1",
     "mime": "2.6.0",
     "minimatch": "^5.1.6",
-    "proxy-agent": "^6.3.1"
+    "proxy-agent": "^6.3.1",
+    "ts-retry-promise": "^0.7.1"
```
shetzel, line 40: Need lock file update as well

mshanemc: nope, there was no change to it.  That's the nature of these--they were working because they were already in the lockfile.  [Until whatever was causing it to be in the lockfile changes.]

cristiand391 (approved):

PR #1219, opened 2024-01-19
fix: use appexchange org for last resort api version

shetzel:
### What does this PR do?
When an API version is needed yet not provided in other ways to the SDR APIs, make a call to the appexchange org to get the most recent API version.  Also, save the result so that all other calls to `getCurrentApiVersion()` on the same process use that value rather than making another call to that endpoint.

@W-14848514@

### DIFF
--- a/src/registry/coverage.ts
+++ b/src/registry/coverage.ts
@@ -9,6 +9,8 @@
 import { OptionsOfTextResponseBody } from 'got';
 import got from 'got';
 import { ProxyAgent } from 'proxy-agent';
+import { isString } from '@salesforce/ts-types';
+import { SfError } from '@salesforce/core';
 import { CoverageObject } from '../../src/registry/types';
 
 const getProxiedOptions = (url: string): OptionsOfTextResponseBody => ({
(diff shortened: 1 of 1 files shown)


src/registry/coverage.ts
```diff
@@ -27,10 +27,15 @@ type ApiVersion = {
...
 
+let apiVer: number;
+
 export const getCurrentApiVersion = async (): Promise<number> => {
-  const apiVersionsUrl = 'https://dx-extended-coverage.my.salesforce-sites.com/services/data';
-  const lastVersionEntry = (await got(getProxiedOptions(apiVersionsUrl)).json<ApiVersion[]>()).pop() as ApiVersion;
-  return +lastVersionEntry.version;
+  if (apiVer === undefined) {
+    const apiVersionsUrl = 'https://appexchange.salesforce.com/services/data';
+    const lastVersionEntry = (await got(getProxiedOptions(apiVersionsUrl)).json<ApiVersion[]>()).pop() as ApiVersion;
```
mshanemc, line 35: have a mild preference for `.at(-1)` over `.pop()` to avoid using the mutating Array methods (fine here, no side effects, just more of "if you never use it, you don't have to think about whether this time is safe").

src/registry/coverage.ts
```diff
@@ -27,10 +27,15 @@ type ApiVersion = {
...
+    apiVer = +lastVersionEntry.version;
+  }
+  return apiVer;
```
mshanemc, line 38: more meta--if this *does* fail (network issues, org down, etc) would the error be meaningful for users to know what's happening?

it's 
```
Uncaught RequestError: getaddrinfo ENOTFOUND appexchange.salesforce.com
```

And I think it'd be nice to tell them why we were doing that and what they might do about it (ex: setting or passing in an api version).  It's hard for SDR to know what they're doing to be more specific.

mshanemc (approved):
QA'd, works fine.

mshanemc (commented):
QA'd, works fine.

src/registry/coverage.ts
shetzel: Internally, the error is never displayed to an end user because if it fails, the code in `ComponentSet` will just use api version 58.0.

Seems this method is exported though which means anyone could call it, and in that case this function has a misleading name and could cause other problems.  I would argue this should not be exported.

mshanemc (approved):

PR #1239, opened 2024-02-19
feat: absolute paths for string replacements for out-of-project

mshanemc:
### What does this PR do?
replacements require reading the project file.  The Project class resolves project using cwd.

DOCe needs to pass in the Project location, so we can rewrite the paths as absolute (relative to the project dir).  
ComponentSetBuilder now accepts a `projectDir` prop on its input, which is the recommended way to do this.  Otherwise, constructing a ComponentSet and setting its `projectDirectory` is also possible.

@W-14970199@


test/nuts/local/replacements/extractZip.ts
```diff
@@ -0,0 +1,27 @@
...
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import * as fs from 'node:fs';
+import * as path from 'node:path';
+import * as JSZip from 'jszip';
+
+export const extractZip = async (zipBuffer: Buffer, extractPath: string) => {
```
mshanemc, line 12: was going to be copy-pasted into 3 files so now it's a function

test/nuts/local/replacements/replacements.nut.ts
```diff
@@ -42,16 +26,7 @@ describe('e2e replacements test', () => {
     // Hack: rewrite the file replacement locations relative to the project
     const projectJsonPath = path.join(session.project.dir, 'sfdx-project.json');
     const original = await fs.promises.readFile(projectJsonPath, 'utf8');
-    await fs.promises.writeFile(
-      projectJsonPath,
-      original
-        // we're putting this in a json file which doesnt like windows backslashes.  The file will require posix paths
```
mshanemc, line 48: the NUT now uses relative filenames for these.  They work using absolute, but I expect that to be less common.

mshanemc:
QA'd by DOCe folks.

PR #1240, opened 2024-02-20
Sm/deterministic-xml-recompose

mshanemc:
### What does this PR do?
when recomposing source => mdapi format, the array of elements (ex: <fields> in CustomObject) are sorted use localCompare.

This ensures consistency in the xml output across OS.

Previously, they would be in the order that fs.readdir would return.


src/convert/transformers/decomposedMetadataTransformer.ts
```diff
@@ -227,14 +228,6 @@ const getDefaultOutput = (component: MetadataComponent): SourcePath => {
   return join(baseComponent.getPackageRelativePath(baseName, 'source'), output);
 };
 
-/** handle wide-open reading of values from elements inside any metadata xml file.
- * Return the value of the matching element if supplied, or defaults `fullName` then `name`  */
-const extractUniqueElementValue = (xml: JsonMap, elementName?: string): string | undefined =>
```
mshanemc, line 232: moved to a utility module so it can be shared between transformers and finalizers

src/convert/convertContext/recompositionFinalizer.ts
```diff
@@ -0,0 +1,184 @@
+/*
```
mshanemc, line 1: this is what needs the most review attention.  It's quite different from the predecessor in the old ConvertContext pile

src/convert/convertContext.ts
```diff
@@ -1,391 +0,0 @@
...
-      const parentObj = parentXmlObj[parentName] as JsonMap;
-
-      if (!parentObj[groupName]) {
-        parentObj[groupName] = [];
-      }
-
-      // it might be an object and not an array.  Example: custom object with a Field property containing a single field
-      const group = ensureArray(parentObj[groupName]) as JsonArray;
-
-      group.push(childContents);
```
mshanemc, line 145: mutation/side effect as the mechanism for setting values.  This took me a while to figure out why pushing to `group` would do anything.

test/snapshot/helper/conversions.ts
```diff
@@ -89,7 +91,8 @@ export const compareTwoXml = (file1: string, file2: string): Chai.Assertion => {
     ignoreDeclaration: true,
     numberParseOptions: { leadingZeros: false, hex: false },
   });
-  return expect(parser.parse(file1)).to.deep.equal(parser.parse(file2));
+
+  return expect(parser.parse(file1)).to.deep.equalInAnyOrder(parser.parse(file2));
```
mshanemc, line 95: when comparing xml files, allow for them to "match" if they have array with the same elements out of order.

This means you can receive wrong-order-xml from retrieve or existing projects.  The decompose-recompose trip will "fix" your order (and that's tested with the snapshot test), but the resulting file should be semantically the same if not literally.

test/snapshot/sampleProjects/customObjects-and-children/__snapshots__/verify-md-files.expected/objects/Broker__c.object
```diff
@@ -106,23 +106,23 @@
         <type>Phone</type>
     </fields>
     <fields>
-        <fullName>Picture_IMG__c</fullName>
+        <fullName>Picture__c</fullName>
```
mshanemc, line 109: fyi, `_` comes early when alphabetizing

src/resolve/sourceComponent.ts
```diff
@@ -224,14 +225,16 @@ export class SourceComponent implements MetadataComponent {
     if (!this.parent) {
       return parentXml;
     }
-    const children = ensureArray(
-      get(parentXml, `${this.parent.type.name}.${this.type.xmlElementName ?? this.type.directoryName}`)
-    ) as T[];
+    // console.log('parent xml', JSON.stringify(parentXml, undefined, 2));
+    // console.log(getXmlElement(this.type));
```
shetzel, line 229: Delete all commented console logs in this file.

shetzel (approved):

PR #1242, opened 2024-02-22
test: mpd-snapshots

mshanemc:
### What does this PR do?
brings project from https://github.com/salesforcecli/sample-project-multiple-packages

makes snapshot assertions about existing behavior
- recompose that project into mdapi format
- apply the mdapi format over existing source with no change

covers tab (basic), object (decomposed), bundle (lwc), label (nonDecomposed)

codifies 2 known bad existing behaviors with nonDecomposed (customLabels)
- the path is not relative to the project (this was already in a labels snapshot test)
- when a mdapi file is decomposed into an existing project, the entire file contents (all labels) are written to each pkgDir that has a customLabels file with a matching name (instead of writing individual labels back to the correct pkgDir)

related to [@W-14608972@](https://gus.lightning.force.com/a07EE00001frGXlYAM)


test/snapshot/helper/conversions.ts
```diff
@@ -60,37 +71,62 @@ export const sourceToMdapi = async (testDir: string): Promise<string[]> => {
...
+const getFullPath = (file: fs.Dirent) => path.join(file.path, file.name);
+
 /** dirEnts are sometimes folder, we don't want those.  And we need the full paths */
-const dirEntsToPaths = (dirEnts: fs.Dirent[]): string[] =>
-  dirEnts.filter((file) => file.isFile()).map((file) => path.join(file.path, file.name));
+export const dirEntsToPaths = (dirEnts: fs.Dirent[]): string[] => dirEnts.filter(isFile).map(getFullPath);
 
 const shouldIgnore = (file: string): boolean => {
   // binary zip/unzip isn't exactly the same, so we "skip" that one
   if (file.includes('leafletjs.resource')) return true;
```
shetzel, line 106: Should there be another utility for `isBinary()` to detect files like this versus hardcoding specific files?

mshanemc: yeah, maybe if this turns out to be pretty common.  

I'm not sure yet if it's "zip/unzip isn't idempotent for **any** binary file" or if it's just "something funny happens with leaflet" because I don't have any other zipped statics in tests yet.

But I like the idea if this is getting out of hand

PR #1261, opened 2024-03-19
fix: allow the same SC in delete and deploy manifests

WillieRuemmele:
### What does this PR do?
allows the same source component to be placed in all types of manifests (pre/post/manifest) simultaneously

@W-15231435@

### Functionality Before

you could not pre-deploy-delete, deploy, or deploy, post-deploy-delete the same MD

### Functionality After

now you can 🚀

### DIFF
--- a/src/collections/componentSet.ts
+++ b/src/collections/componentSet.ts
@@ -26,19 +26,19 @@ import {
 import { XML_DECL, XML_NS_KEY, XML_NS_URL } from '../common';
 import {
   ComponentLike,
+  ConnectionResolver,
   ManifestResolver,
   MetadataComponent,
   MetadataMember,
   MetadataResolver,
-  ConnectionResolver,
   SourceComponent,
 } from '../resolve';
 import { getCurrentApiVersion, MetadataType, RegistryAccess } from '../registry';
 import {
   DestructiveChangesType,
+  FromConnectionOptions,
   FromManifestOptions,
   FromSourceOptions,
-  FromConnectionOptions,
   PackageManifestObject,
   PackageTypeMembers,
 } from './types';
@@ -82,7 +82,7 @@ export class ComponentSet extends LazyCollection<MetadataComponent> {
   public fullName?: string;
   public forceIgnoredPaths?: Set<string>;
   private logger: Logger;
-  private registry: RegistryAccess;
+  private readonly registry: RegistryAccess;
   private components = new DecodeableMap<string, DecodeableMap<string, SourceComponent>>();
 
   // internal component maps used by this.getObject() when building manifests.
@@ -436,7 +436,7 @@ export class ComponentSet extends LazyCollection<MetadataComponent> {
       typeMembers.push({ members: members.sort(), name: typeName });
     }
 
-    const pkg = {
+    return {
       Package: {
         ...{
           types: typeMembers.sort((a, b) => (a.name > b.name ? 1 : -1)),
@@ -445,8 +445,6 @@ export class ComponentSet extends LazyCollection<MetadataComponent> {
         ...(this.fullName ? { fullName: this.fullName } : {}),
       },
     };
-
-    return pkg;
   }
 
   /**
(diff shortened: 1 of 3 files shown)


src/collections/componentSet.ts
```diff
@@ -721,7 +739,7 @@ export class ComponentSet extends LazyCollection<MetadataComponent> {
 
 const sourceKey = (component: SourceComponent): string => {
   const { fullName, type, xml, content } = component;
-  return `${type.name}${fullName}${xml ?? ''}${content ?? ''}`;
+  return `${type.name}${fullName}${xml ?? ''}${content ?? ''}${component.isMarkedForDelete()}`;
```
shetzel, line 742: If this really is needed I think it should be more descriptive than appending `false` or `true`.  I'm hopeful that we can rely on the source components being accurate and assigning them into the correct component sets on add.

src/collections/componentSet.ts
```diff
@@ -510,32 +520,16 @@ export class ComponentSet extends LazyCollection<MetadataComponent> {
...
       component.setMarkedForDelete(deletionType);
       this.logger.debug(`Marking component for delete: ${component.fullName}`);
-      const deletions = this.destructiveComponents[deletionType];
-      if (!deletions.has(key)) {
-        deletions.set(key, new DecodeableMap<string, SourceComponent>());
+      if (!this.destructiveComponents[deletionType].has(key)) {
+        this.destructiveComponents[deletionType].set(key, new DecodeableMap<string, SourceComponent>());
       }
-      deletions.get(key)?.set(sourceKey(component), component);
+      this.destructiveComponents[deletionType].get(key)?.set(sourceKey(component), component);
```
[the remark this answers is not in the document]
WillieRuemmele: but it doesn't end up updating anything of importance the old way.

shetzel:
QA:
`package-classes.xml` -> contains all classes in the project
`preDestructiveChanges-class.xml` -> contains 1 class; Foo.cls which also exists in the project
`preDestructiveChanges-classes.xml` -> contains 2 classes; Foo.cls and Bar.cls which also exist in the project
`package-empty.xml -> no types listed

1. `project deploy start --manifest package-classes.xml --pre-destructive-changes preDestructiveChanges-class.xml` -> deletes a class (Foo) in pre, deploys all classes including Foo.  ✅   Output does not explicitly display deleted class Foo.
2. `project deploy start --manifest package-empty.xml --pre-destructive-changes preDestructiveChanges-class.xml` -> deletes a class (Foo) in pre.  ✅  Output lists that the class was deleted in a deleted section and also was deleted in the deploy section, which is likely from PDR.
3. `project deploy start --manifest preDestructiveChanges-class.xml --pre-destructive-changes preDestructiveChanges-class.xml` -> deletes a class (Foo) in pre, then deploys that same class.  ✅  Output does not explicitly display deleted class Foo.
4. `project deploy start --manifest package-classes.xml --pre-destructive-changes preDestructiveChanges-classes.xml` -> deletes classes (Foo and Bar) in pre, deploys all classes including Foo and Bar.  ✅   Output does not explicitly display deleted classes Foo and Bar.
5. `project deploy start --manifest package-classes.xml --pre-destructive-changes package-empty.xml` -> nothing deleted.  all classes deployed.  ✅

src/resolve/treeContainers.ts
```diff
@@ -11,8 +11,8 @@ import { statSync, existsSync, readdirSync, createReadStream, readFileSync } fro
 import * as JSZip from 'jszip';
 import { Messages, SfError } from '@salesforce/core';
 import { isString } from '@salesforce/ts-types';
-import { baseName, parseMetadataXml } from '../utils/path';
-import type { SourcePath } from '../common/types';
+import { baseName, parseMetadataXml } from '../utils';
```
mshanemc, line 14: lets only load the required files instead of the barrel

src/resolve/sourceComponent.ts
```diff
@@ -259,10 +259,8 @@ export class SourceComponent implements MetadataComponent {
       delete this.destructiveChangesType;
     } else {
       this.markedForDelete = true;
-      // eslint-disable-next-line no-unused-expressions
-      destructiveChangeType === DestructiveChangesType.PRE
-        ? (this.destructiveChangesType = DestructiveChangesType.PRE)
-        : (this.destructiveChangesType = DestructiveChangesType.POST);
+      this.destructiveChangesType =
```
mshanemc, line 262: I had to stare at this one for a second.  comment could be
// [POST, true, undefined ] => POST.  Only PRE means PRE

mshanemc (approved):

PR #1262, opened 2024-03-19
fix: refactor for better typing and readability

shetzel:
### What does this PR do?
Adds 502, 503, and 420 error codes to list of responses to retry during deploy polling.

@W-15289218@

### DIFF
--- a/src/client/metadataTransfer.ts
+++ b/src/client/metadataTransfer.ts
@@ -200,6 +200,34 @@ export abstract class MetadataTransfer<
     return getConnectionNoHigherThanOrgAllows(this.usernameOrConnection, this.apiVersion);
   }
 
+  // eslint-disable-next-line class-methods-use-this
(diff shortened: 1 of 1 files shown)


src/client/metadataTransfer.ts
```diff
@@ -229,23 +229,34 @@ export abstract class MetadataTransfer<
           await Lifecycle.getInstance().emitWarning('Metadata API response not parseable');
           return { completed: false };
         }
+        const retryableErrors = [
```
mshanemc, line 232: should work.  I think this has gotten big/complex enough to make it it separate function (SRP) vs. being inside `poll` 

`isRetryable(error: unknown) : boolean` and then let it contain the error list and manage the type checks on what it's being passed.

mshanemc (approved):

mshanemc (approved):

PR #1265, opened 2024-03-28
fix: non-zip conversions return empty arrays for converted and deleted

mshanemc:
[skip-validate-pr]

restore a unspecified type behavior that packaging was relying on


WillieRuemmele (approved):
✅ : linked into `packaging@main` , it built and passed UTs

PR #1271, opened 2024-04-01
test: workflow fix confirmation

mshanemc:
### What does this PR do?

1. use customer-supplied repo from @amtrack to validate behavior of conversion to mdapi when using mpd
1. remove one top-level file that got left laying around from previous PRs

https://github.com/forcedotcom/cli/discussions/2356, @W-15379166@


test/snapshot/sampleProjects/preset-workflow/__snapshots__/verify-source-files.expected/force-app/main/default/workflows/Account.workflow-meta.xml
```diff
@@ -1,51 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
mshanemc, line 1: this file should not have existed after decomposing

PR #1273, opened 2024-04-02
Wr/add missing local components

WillieRuemmele:

#https://github.com/forcedotcom/cli/issues/2603, @W-14675377@

### Functionality Before

`getFileResponses` would only return files that existed locally

### Functionality After

`getFileResponses` now returns all results from the server, with a special note for ones not found locally


src/client/metadataApiDeploy.ts
```diff
@@ -370,6 +374,46 @@ const deleteNotFoundToFileResponses =
           : [];
       });
 
+const serverResponseNotFoundLocally =
+  (cs: ComponentSet) =>
+  (messageMap: Map<string, DeployMessage[]>): FileResponse[] => {
+    const sourceKeys = cs.getSourceComponents().toArray().map(toKey);
```
mshanemc, line 380: for a large deploy, this'll be a bit more performant if you make sourceKeys a `set` and use `.has(k)`

shetzel (approved):

shetzel:
QA:
With changes, there is now a warning message (e.g.): `Warning: CustomField, Property__c.City2__c, returned from org, but not found in the local project`
Without changes, there is no warning message and no file listed in the deployed files section.

PR #1274, opened 2024-04-02
Mz/integrate core bundle

mingxuanzhangsfdx:
### What does this PR do?
This PR creates a workflow to publish sdr-bundle package in order to be consumed by source-tracking-bundle and extensions

@W-15288768@

### Functionality Before

There is no such workflow before

### Functionality After

The workflow is created.

### DIFF
--- a/.github/workflows/releaseWithCoreBundle.yml
+++ b/.github/workflows/releaseWithCoreBundle.yml
@@ -1,0 +1,16 @@
+name: publish source-deploy-retrieve-bundle
+on:
+  workflow_dispatch:
+    inputs:
+      branch:
+        description: 'Set the branch to use for release'
+        type: string
+        required: false
+        default: 'main'
+
+jobs:
+  call-release-workflow:
+    uses: forcedotcom/bundle-publish-scripts/.github/workflows/releaseWithCoreBundle.yml@main
+    secrets: inherit
+    with:
+      branch: ${{ inputs.branch }}
--- a/package.json
+++ b/package.json
@@ -202,5 +202,8 @@
       ],
       "output": []
     }
+  },
+  "publishConfig": {
+    "access": "public"
   }
 }


gbockus-sf (approved):
LGTM 🚢

.github/workflows/releaseWithCoreBundle.yml
```diff
@@ -0,0 +1,16 @@
+name: publish source-deploy-retrieve-bundle
+on:
+  workflow_dispatch:
```
mshanemc, line 3: so you're going to manually push the button every time, or is some other automated process is going to wait for it?

Suggestion: maybe this should run after each time something else runs (ex: after a gh release publishes, or after the normal npm publish action runs) so that you have on for every version?

Or is this just temporary and you'll automate more later?

.github/workflows/releaseWithCoreBundle.yml
```diff
@@ -0,0 +1,16 @@
...
+    inputs:
```
mshanemc, line 4: you might also want to parameterize that npm token in case you ever have something outside `forcedotcom` that wants to use your workflows.

It's a bit more useful than `secrets: inherit` and then keeps all the secrets out that it doesn't need.

.github/workflows/releaseWithCoreBundle.yml
```diff
@@ -0,0 +1,16 @@
...
+      branch:
+        description: 'Set the branch to use for release'
+        type: string
+        required: false
+        default: 'main'
+
+jobs:
+  call-release-workflow:
+    uses: forcedotcom/bundle-publish-scripts/.github/workflows/releaseWithCoreBundle.yml@main
```
mshanemc, line 13: can I introduce you to fs.readdir with recursive:true instead of writing yet another one of these?  https://github.com/forcedotcom/bundle-publish-scripts/blob/4c0c4a4faf393f7c585c0bf8da153592ecbbee98/scripts/updateForCoreBundle.js#L79.

There's also a `withFileTypes` option if you want to filter out dirs or pass a filter function.

mshanemc (approved):

.github/workflows/releaseWithCoreBundle.yml
mingxuanzhangsfdx: Thanks I just had a quick try and the new function only updates the first file. Not sure what went wrong, and I will try to dig into it later

PR #1275, opened 2024-04-03
fix: expand mdTransfer error

cristiand391:
### What does this PR do?

updates `MetadataTransfer.pollStatus` error handling to expand the mdTransfer error with error data/actions if the failure throws an SfError.

An example of this is STL throwing on source-tracking after a CLI deploy (which happens on the postdeploy hook on SDR).
One customer is getting the `multiple git error` iso-git failure on STL but the path errors don't show up in the CLI final error.
See this thread:
https://salesforce-internal.slack.com/archives/G02K6C90RBJ/p1711480148180689

@W-15347866@

### Functionality Before

errors in post retrieve/deploy get swallowed in mdTransfer catch block.

### Functionality After

mdTransfer will include cause error data (if instanceof SfError) and actions.

<insert gif and/or summary>

### DIFF
--- a/src/client/metadataTransfer.ts
+++ b/src/client/metadataTransfer.ts
@@ -115,14 +115,26 @@ export abstract class MetadataTransfer<
       }
       return result;
     } catch (e) {
-      const err = e as Error;
-      const error = new SfError(messages.getMessage('md_request_fail', [err.message]), 'MetadataTransferError');
-      error.setData({
-        id: this.id,
-      });
+      const err = e as Error | SfError;
+      const error = new SfError(messages.getMessage('md_request_fail', [err.message]), 'MetadataTransferError', err);
+
       if (error.stack && err.stack) {
         // append the original stack to this new error
         error.stack += `\nDUE TO:\n${err.stack}`;
+
+        if (err instanceof SfError && err.data) {
+          // this keeps SfError data for failures in post deploy/retrieve.
+          error.setData({
+            id: this.id,
+            causeErrorData: error.data,
+          });
+
+          error.actions = err.actions;
+        } else {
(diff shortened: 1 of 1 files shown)


src/client/metadataTransfer.ts
```diff
@@ -115,14 +115,26 @@ export abstract class MetadataTransfer<
...
       return result;
     } catch (e) {
-      const err = e as Error;
+      const err = e as Error | SfError;
       const error = new SfError(messages.getMessage('md_request_fail', [err.message]), 'MetadataTransferError');
-      error.setData({
-        id: this.id,
-      });
+
       if (error.stack && err.stack) {
```
mshanemc, line 121: how would `error.stack` ever be truthy if we just constructed it?

src/client/metadataTransfer.ts
```diff
@@ -115,14 +115,26 @@ export abstract class MetadataTransfer<
       }
       return result;
     } catch (e) {
```
mshanemc, line 117: this might be better as an exported function (pass in the `e` and the `id`) so that we could have a UT that imports it and throws various errors at it (regular Error, SfError, an `isogit` error) and asserts that we're giving back SfErrors with the correct input.

mshanemc (changes requested):

src/client/metadataTransfer.ts
cristiand391: all errors have a stack trace set when instantiated:
```
node -e  'const err= new Error(); console.log(err.stack);'
```

Screenshot 2024-04-03 at 6 23 26 PM

mshanemc (approved):

PR #1294, opened 2024-04-22
feat: allow destructive changes with MetadataOptions

shetzel:
### What does this PR do?
Allows defining metadata type/name pairs to be deleted (pre or post) as part of a metadata deploy.  Adds snapshot tests for this behavior.

@W-15306474@


test/snapshot/sampleProjects/destructiveCompSets/snapshots.test.ts
```diff
@@ -0,0 +1,159 @@
...
+    });
+
+    const convertedFiles = await getConvertedFilePaths(postTestOutputDir);
+    for (const file of convertedFiles) {
+      await fileSnap(file, testDir);
+    }
+    dirsAreIdentical(path.join(snapshotsDir, 'testOutput', 'post1'), postTestOutputDir);
+  });
+
+  it('2 post-destructive and 2 deployments', async () => {
```
WillieRuemmele, line 113: should/could we add a test covering 1+ pre or post destructive changes, without any constructive? Seems like it'll be working - but would be good to have for future changes

WillieRuemmele: or maybe we don't want to support that?  if the CLIs not adding new flags to use this, I'm not sure if other SDR consumers would like that ability

WillieRuemmele (approved):

test/snapshot/sampleProjects/destructiveCompSets/snapshots.test.ts
shetzel: I can add a test for those cases.  👍

WillieRuemmele:
### QA Notes
> since there's no flags, or exposure through the CLI, I built a small script to test the behavior

```ts
    const c = await ComponentSetBuilder.build({metadata:{
      directoryPaths:['force-app/main/default/classes'],
        destructiveEntriesPost: ['ApexClass:myClassPOST'],
        destructiveEntriesPre: ['ApexClass:myClassPRE'],
        metadataEntries: ['ApexClass:myClass']
      }})

    const p = await c.getPackageXml(4, DestructiveChangesType.PRE);
    const x = await c.getPackageXml(4, DestructiveChangesType.POST);

    // eslint-disable-next-line no-console
    console.log(p,x);
```

✅ : I ensured that both variables, `p`/`x` are correct.
✅ : when deployed, the classes were deleted

PR #1297, opened 2024-04-25
fix: md names handle spaces, colons, and both

mshanemc:
### What does this PR do?
handle a devOpsCenter followup to @W-15253267@


EstebanRomero84 (approved):
Looks good, thank you

PR #1307, opened 2024-05-08
fix: use `projectDir` when getting API version

cristiand391:
### What does this PR do?

Pass `this.projectDirectory` to `SfProject.resolve` when resolving the manifest API version from `sfdx-project` for a deploy.
This was the last place where SDR tries to resolve the sfdx project from `process.cwd`, which fails for consumers running code outside of a project (vscode/devops center).

@W-12488179@
@W-15734097@

### Functionality Before

if you didn't specify `sourceapiversion` when building a component set, on a deploy it would do an upward dir search for `sfdx-project.json` starting from `process.cwd` just to read the API version from it.
For code not running from the project it would throw "couldn't find the project" err even if you passed the `projectDir` to the component set.

### Functionality After

the project will be resolved from `projectDir` if available.

`sf` CLI was already doing this so we shouldn't see any behavior change:
https://github.com/salesforcecli/plugin-deploy-retrieve/blob/3417931c3fbbe31d76a7875bbed372529c9a1cbd/src/utils/deploy.ts#L94
https://github.com/salesforcecli/plugin-deploy-retrieve/blob/3417931c3fbbe31d76a7875bbed372529c9a1cbd/src/utils/project.ts#L16


cristiand391:
I updated the snapshots rather than setting the sourceApiVersion when building the component set for the tests so that snap tests also fail if this change.

mshanemc (approved):

PR #1308, opened 2024-05-10
fix: prevent empty parent xml overwrites

mshanemc:
### What does this PR do?
when retrieving a decomposed type (ex a customField) the API returns the parent (customObject) with only the requested field inside it.
Decomposition removes the requested field from the parent, leaving it empty (a CustomObject with no child xml nodes) and then writes that.

https://github.com/forcedotcom/cli/issues/2865
@W-15727064@

### repro
have a repo with custom objects in it.  (dreamhouse works fine)
`sf project retrieve start -m CustomField:Foo__c.Bar__c`

before: the object-meta.xml would be blank unless you retrieve the CustomObject, too
after: the object-meta.xml keeps its original values

Note: you can still retrieve a field and end up with an empty parent, as long as you didn't have the parent before.  To get the parent's contents, include it in your retrieve (ex:  `sf project retrieve start -m CustomField:Foo__c.Bar__c -m CustomObject:Foo__c`)


mdonnalley:
QA:

**setup**
- checkout branch, `yarn && yarn build`
- yarn link `@salesforce/source-deploy-retrieve` in PDR
- `sf plugins link --no-install` in PDR

In dreamhouse-lwc
`sf project retrieve start -m CustomField:Broker__c.Broker_Id__c`

🟢 `force-app/main/default/objects/Broker__c/Broker__c.object-meta.xml` is not empty

mdonnalley (approved):

PR #1321, opened 2024-05-21
fix: components marked for deletion don't go into the include filter

mshanemc:
### What does this PR do?
componentSet.fromSource will take the passed in `include` property, then filter out any sourceComponents that are markedForDelete before applying that filter to the tree.

https://github.com/forcedotcom/cli/issues/2826 
@W-15771541@

### DIFF
--- a/src/collections/componentSet.ts
+++ b/src/collections/componentSet.ts
@@ -73,10 +73,10 @@ export class ComponentSet extends LazyCollection<MetadataComponent> {
    * When not present, sfdx-core's SfProject will use the current working directory.
    */
   public projectDirectory?: string;
+  public readonly registry: RegistryAccess;
   public fullName?: string;
   public forceIgnoredPaths?: Set<string>;
   private logger: Logger;
-  private readonly registry: RegistryAccess;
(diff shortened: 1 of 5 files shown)


WillieRuemmele (approved):

src/collections/componentSet.ts
```diff
@@ -750,3 +740,24 @@ const splitOnFirstDelimiter = (input: string): [string, string] => {
...
+    return given.include ? { ...given, include: removeMarkedForDelete(given.include) } : given;
+  } else {
+    return { fsPaths: [given] };
+  }
+};
+
+/** remove components that are marked for delete */
+const removeMarkedForDelete = (include: ComponentSet): ComponentSet => {
+  const filtered = new ComponentSet([], include.registry);
+  for (const component of include) {
```
jayree, line 757: @mshanemc you could also use `include.getSourceComponents()` instead of `component instanceof SourceComponent`

mshanemc:
abandoning, will try to get https://github.com/forcedotcom/source-deploy-retrieve/pull/1328 to work

PR #1329, opened 2024-05-29
Wr/improve did you mean

WillieRuemmele:
### What does this PR do?

fixes a bug when a MD format file was converted to MD format, it would strip the last char from the file path, and then cause a error when trying to identify the metadata with the then incorrect suffix

fixes error
it now emits a warning when md format files are found and the `toFormat` is metadata

@W-15783171@

### Functionality Before

```bash
 ➜  sf project convert source --output-dir md --source-dir force-app/main/default/enablementMeasureDefinitions
Error (1): /Users/william.ruemmele/projects/scratches/dreamhouse-lwc/md/enablementMeasureDefinitions/testing.enablementMeasureDefinitio: Could not infer a metadata type

Try this:

A metadata type lookup for "testing.enablementMeasureDefinitio" found the following close matches:
-- Did you mean ".enablementMeasureDefinition" instead for the "EnablementMeasureDefinition" metadata type?
```
### Functionality After

```bash
 ➜  ../../oss/plugin-deploy-retrieve/bin/run.js project convert source --output-dir md/format --source-dir force-app/main/default/enablementMeasureDefinitions 
Warning: Potential metadata-format file enablementMeasureDefinitions/testing.enablementMeasureDefinition found in a source-format directory
Source was successfully converted to Metadata API format and written to the location: /Users/william.ruemmele/projects/scratches/dreamhouse-lwc/md/format
```

### DIFF
--- a/test/nuts/suggestType/suggestType.nut.ts
+++ b/test/nuts/suggestType/suggestType.nut.ts
@@ -92,10 +92,10 @@ describe('suggest types', () => {
     });
     expect(lifecycleSpy.calledOnce).to.be.true;
     expect(lifecycleSpy.args[0][0]).to.equal(
-      `Potential metadata-format file ${join(
+      `Found a file (${join(
         'enablementMeasureDefinitions',
         'measure.enablementMeasureDefinition'
-      )} found in a source-format directory`
+      )}) that appears to be in metadata format, but the directory it's in is for source formatted files.`
     );
 
     fs.rmSync('output', { recursive: true, force: true });
--- a/test/nuts/suggestType/suggestType.nut.ts
+++ b/test/nuts/suggestType/suggestType.nut.ts
@@ -90,8 +90,7 @@ describe('suggest types', () => {
       genUniqueDir: false,
       outputDirectory: 'output',
     });
-    expect(lifecycleSpy.callCount).to.equal(1);
-    expect(lifecycleSpy.args.flat()).to.deep.equal([
+    expect(lifecycleSpy.args.flat()).to.deep.include([
       `Found a file (${join(
         'enablementMeasureDefinitions',
         'measure.enablementMeasureDefinition'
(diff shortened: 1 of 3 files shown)


test/nuts/suggestType/suggestType.nut.ts
```diff
@@ -66,6 +69,38 @@ describe('suggest types', () => {
...
+        ),
+      ],
+    });
+    const converter = new MetadataConverter();
+    await converter.convert(set, 'metadata', {
+      type: 'directory',
+      genUniqueDir: false,
+      outputDirectory: 'output',
+    });
+    expect(lifecycleSpy.calledOnce).to.be.true;
```
mshanemc, line 93: if this comes back `false` does that mean it was not called, or called twice?
[see below comment...does the warning count matter, or does the test only care that the expected warning happened?]

```suggestion
    expect(lifecycleSpy.callCount).to.equal(1);
```

test/nuts/suggestType/suggestType.nut.ts
```diff
@@ -66,6 +69,38 @@ describe('suggest types', () => {
...
+    expect(lifecycleSpy.args[0][0]).to.equal(
```
mshanemc, line 94: can you write this to assert that the expected error exists, not its ordinality, so that some future warning doesn't break the UT?

mshanemc:
QA:

converted project to MDAPI, deleted applications/dreamhouse, then put the mdapi version in its place
✅ sees warning

change `Broker__c.object-meta.xml` => `Broker__c.object`
📓  no warning (maybe ok, we're only addressing the "basic" type since object is quite a different format?

change a flexipage
✅ get 2 warnings

tested:

✅  apexClass ( w/ content type because those have 2 files, one w/ and without the -meta.xml) 
✅  bundles (lwc, aura) because  bundles don't have a -meta.xml on most of their files in content

PR #1334, opened 2024-06-05
chore: add automation of release and test workflow

mingxuanzhangsfdx:
### What does this PR do?
Automate the release and test with core-bundle

@W-15634780@

### Functionality Before

No automation test against core-bundle and there is only manual release of sdr-bundle

### Functionality After

Automation of the release and test of sdr-bundle


CristiCanizales (approved):
LGTM!

PR #1346, opened 2024-06-19
test: only run the validationSourceDir once (avoid duplicate expects folders)

mshanemc:
[skip-validate-pr]


WillieRuemmele:
✅ : before
```bash
 ➜  ls test/snapshot/sampleProjects/singleCustomFieldRetrieve/__snapshots__/
verify-force-app.expected                     will-not-overwrite-object-meta-xml.expected.1
will-not-overwrite-object-meta-xml.expected   will-retrieve-using-Retrieve.expected
```

after:
```bash
 ➜  ls test/snapshot/sampleProjects/singleCustomFieldRetrieve/__snapshots__/
verify-force-app.expected                   will-retrieve-using-Retrieve.expected
will-not-overwrite-object-meta-xml.expected
```

PR #1347, opened 2024-06-19
fix: remove xml cache

mshanemc:
### What does this PR do?
module-scope cache causes problems for long-lived IDE usage.  this just removes it (goal is to compare the perf tests for EDA, which has a lot of decomposed stuff, to see if it matters at all)

@W-15952068@

### DIFF
--- a/src/convert/convertContext/recompositionFinalizer.ts
+++ b/src/convert/convertContext/recompositionFinalizer.ts
@@ -20,8 +20,6 @@ import { ConvertTransactionFinalizer } from './transactionFinalizer';
 Messages.importMessagesDirectory(__dirname);
 const messages = Messages.loadMessages('@salesforce/source-deploy-retrieve', 'sdr');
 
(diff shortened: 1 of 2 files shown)


shetzel (commented):
the xml cache was specifically for CustomLabels.  Create a project with 2000 CustomLabels and do a convert.  You'll see the perf difference.  With each individual custom label the entire CustomLabels file is re-read so performance, especially on windows, is pretty terrible without the cache.

mshanemc:
> the xml cache was specifically for CustomLabels. Create a project with 2000 CustomLabels and do a convert. You'll see the perf difference. With each individual custom label the entire CustomLabels file is re-read so performance, especially on windows, is pretty terrible without the cache.

the perf tests agreed with you 😄 .  I wanted to see.  Closed in favor of https://github.com/forcedotcom/source-deploy-retrieve/pull/1348

PR #1348, opened 2024-06-19
fix: class-scoped xml cache for recomposition

mshanemc:
### What does this PR do?

multiple converters could be going on simultaneously, or as a long-lived process.  This cache hung around between converts and caused a bug for IDE.

see slack thread https://salesforce-internal.slack.com/archives/G0213NY8PMH/p1718754058826369

removing the cache was quite bad for perf https://github.com/forcedotcom/source-deploy-retrieve/actions/runs/9587973660

@W-15952068@


mingxuanzhangsfdx (approved):
Thanks for the change, it looks good to me.

PR #1356, opened 2024-06-25
fix(perf): fewer adapters

mshanemc:
### What does this PR do?
while doing some background research for W-15771541 (unrelated) I noticed we've got a class (SourceAdapterFactory) that constructs classes (adapters) to answer a question where the answer is stored on the Adapter.

```
public allowMetadataWithContent(): boolean {
    return this.metadataWithContent;
  }
```

since type **determines** Adapter, why not use the type and not have to mess with the adapter.

metadataResolver constructs the factoy to construct adapters, and then might not even use it based on some other conditionals.  Those now check first for early exit.


shetzel (approved):

mshanemc:
closing in favor of #1382

PR #1358, opened 2024-06-27
feat: guarantee fileResponseSuccess has a path

mshanemc:
> really a type refactor; no intended functionality change for SDR

SDR is the source of FileResponse (and its subtype FileResponseSuccess).

All the real code that produces them provide a filepath, but the type has it as `?`.

consumers with strict TS have to do checks for no reason.

fixes part of https://github.com/forcedotcom/eslint-config-salesforce-typescript/actions/runs/9703130477/job/26780562510?pr=246#step:9:9

so I guess it's related to [@W-16116290@](https://gus.lightning.force.com/a07EE00001vThmcYAC)

### DIFF
--- a/test/client/metadataApiDeploy.test.ts
+++ b/test/client/metadataApiDeploy.test.ts
@@ -7,7 +7,7 @@
 import { basename, join, sep } from 'node:path';
 import { MockTestOrgData, TestContext } from '@salesforce/core/testSetup';
 import chai, { assert, expect } from 'chai';
-import { AnyJson, getString } from '@salesforce/ts-types';
+import { AnyJson, ensureString, getString } from '@salesforce/ts-types';
 import { Lifecycle, Messages, PollingClient, StatusResult } from '@salesforce/core';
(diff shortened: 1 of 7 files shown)


test/client/metadataApiDeploy.test.ts
```diff
@@ -548,13 +548,13 @@ describe('MetadataApiDeploy', () => {
             fullName,
             type: type.name,
             state: ComponentStatus.Changed,
-            filePath: content,
+            filePath: ensureString(content),
```
mshanemc, line 551: there's a bunch of these in the test.  The mocks come from SourceComponent, which has `xml` and `content` as optional properties.

the tests assert that they're there.

It's harder to work with the SourceComponent class (ex: to make SourceComponentWithXmlAndContent, etc) the way you would a simple object/type or to put a type guard around it  
...
so it's probably simpler to do this `ensure` stuff for UT.

shetzel (approved):

PR #1359, opened 2024-06-28
chore: use import rather than fs read

mingxuanzhangsfdx:
### What does this PR do?
the unspecified imports or file reads makes esbuild unable to bundle the modules. This PR fixes the issue that the presets json files are not bundled.

@W-16114122@

### Functionality Before
presets json modules are not able to be bundled without specifications, and results in the failure in extensions.

### Functionality After
presets json modules are able to be bundled directly.

### DIFF
--- a/src/registry/variants.ts
+++ b/src/registry/variants.ts
@@ -25,7 +25,7 @@ export type RegistryLoadInput = {
 export const getEffectiveRegistry = (input?: RegistryLoadInput): MetadataRegistry =>
   deepFreeze(firstLevelMerge(registryData as MetadataRegistry, loadVariants(input)));
 
-const presetsJsonMap: Map<string, string> = new Map();
+const presetsJsonMap: Map<string, MetadataRegistry> = new Map();
 
 /** read the project to get additional registry customizations and sourceBehaviorOptions */
 const loadVariants = ({ projectDir }: RegistryLoadInput = {}): MetadataRegistry => {
@@ -81,16 +81,15 @@ const maybeGetProject = (projectDir?: string): SfProjectJson | undefined => {
 };
 
 const loadPresetJson = (): void => {
-  presetsJsonMap.set('decomposeCustomLabelsBeta', JSON.stringify(decomposeCustomLabelsBeta));
-  presetsJsonMap.set('decomposePermissionSetBeta', JSON.stringify(decomposePermissionSetBeta));
-  presetsJsonMap.set('decomposeSharingRulesBeta', JSON.stringify(decomposeSharingRulesBeta));
-  presetsJsonMap.set('decomposeWorkflowBeta', JSON.stringify(decomposeWorkflowBeta));
+  presetsJsonMap.set('decomposeCustomLabelsBeta', decomposeCustomLabelsBeta as MetadataRegistry);
+  presetsJsonMap.set('decomposePermissionSetBeta', decomposePermissionSetBeta as MetadataRegistry);
+  presetsJsonMap.set('decomposeSharingRulesBeta', decomposeSharingRulesBeta as MetadataRegistry);
+  presetsJsonMap.set('decomposeWorkflowBeta', decomposeWorkflowBeta as MetadataRegistry);
 };
 
 const loadPreset = (preset: string): MetadataRegistry => {
   try {
-    const rawPreset = presetsJsonMap.get(preset);
-    return JSON.parse(rawPreset as string) as MetadataRegistry;
+    return presetsJsonMap.get(preset) as MetadataRegistry;
   } catch (e) {
     throw new Error(`Failed to load preset ${preset}. The value is invalid.`);
   }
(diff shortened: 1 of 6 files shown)


src/registry/variants.ts
```diff
@@ -4,12 +4,14 @@
  * Licensed under the BSD 3-Clause license.
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
-import * as fs from 'node:fs';
-import * as path from 'node:path';
 import { Logger, SfProject, SfProjectJson, Lifecycle } from '@salesforce/core';
 import { deepFreeze } from '../utils/collections';
 import { MetadataRegistry } from './types';
 import * as registryData from './metadataRegistry.json';
+import decomposeCustomLabelsBeta from './presets/decomposeCustomLabelsBeta.json';
```
mingxuanzhangsfdx, line 11: I do not use asynchronous imports for the presets jsons. Make the imports async will results in making all outer functions async.

mshanemc: why wouldn't you use the `import * as ` and `as MetadataRegistry` pattern that works for `metadataRegistry.json` to make a Map<string, MetadataRegistry>

turning the json into a string and then turning it back into json seems odd...is that something necessary for the bundling to work?  Why is it not necessary for `metadataRegistry.json`

mingxuanzhangsfdx: I tried to compare the content
``` typescript
import * as path from 'node:path';
import * as fs from 'node:fs';
const pathToCheck = path.join(__dirname, 'presets', 'decomposeCustomLabelsBeta.json');
console.log(fs.readFileSync(pathToCheck, 'utf-8'))
import * as decomposeCustomLabelsBeta from './presets/decomposeCustomLabelsBeta.json';
console.log(JSON.stringify(decomposeCustomLabelsBeta));
```

And noticed the output of `JSON.stringify(decomposeCustomLabelsBeta)` is not what I want. 

There is a dummy default field there, which makes it different from the output of `fs.readFileSync(pathToCheck, 'utf-8')`

``` json
{"childTypes":{"customlabel":"customlabels"},"suffixes":{"label":"customlabel"},"strictDirectoryNames":{"labels":"customlabels"},"types":{"customlabels":{"children":{"suffixes":{"label":"customlabel"},"types":{"customlabel":{"directoryName":"test","id":"customlabel","name":"CustomLabel","suffix":"label","isAddressable":false,"supportsWildcardAndName":true,"uniqueIdElement":"fullName","xmlElementName":"labels"}}},"strictDirectoryName":true,"directoryName":"labels","id":"customlabels","ignoreParsedFullName":false,"name":"CustomLabels","strategies":{"adapter":"decomposed","decomposition":"topLevel","transformer":"decomposed"},"suffix":"labels","supportsPartialDelete":true}},"default":{"childTypes":{"customlabel":"customlabels"},"suffixes":{"label":"customlabel"},"strictDirectoryNames":{"labels":"customlabels"},"types":{"customlabels":{"children":{"suffixes":{"label":"customlabel"},"types":{"customlabel":{"directoryName":"test","id":"customlabel","name":"CustomLabel","suffix":"label","isAddressable":false,"supportsWildcardAndName":true,"uniqueIdElement":"fullName","xmlElementName":"labels"}}},"strictDirectoryName":true,"directoryName":"labels","id":"customlabels","ignoreParsedFullName":false,"name":"CustomLabels","strategies":{"adapter":"decomposed","decomposition":"topLevel","transformer":"decomposed"},"suffix":"labels","supportsPartialDelete":true}}}}
```

mshanemc: that's an artifact of import/export (like `export default foo` it's there for you in case you're importing it not using the `*`)

src/registry/variants.ts
```diff
@@ -75,14 +80,18 @@ const maybeGetProject = (projectDir?: string): SfProjectJson | undefined => {
...
+  presetsJsonMap.set('decomposePermissionSetBeta', decomposePermissionSetBeta as MetadataRegistry);
+  presetsJsonMap.set('decomposeSharingRulesBeta', decomposeSharingRulesBeta as MetadataRegistry);
+  presetsJsonMap.set('decomposeWorkflowBeta', decomposeWorkflowBeta as MetadataRegistry);
+};
 
+const loadPreset = (preset: string): MetadataRegistry => {
   try {
-    const rawPreset = fs.readFileSync(pathToCheck, 'utf-8');
-    return JSON.parse(rawPreset) as MetadataRegistry;
+    return presetsJsonMap.get(preset) as MetadataRegistry;
```
mshanemc, line 92: fyi, `Map.get` returns `undefined` when you ask for something not in the map.  It doesn't throw an error

mshanemc: TS would have told you had a bug, but using ` as` hides the error from yourself

mshanemc (approved):

PR #1368, opened 2024-07-15
fix: only read multiple custom labels files once during conversion

shetzel:
When multiple custom labels files are in a project, only read them once during conversion.  There was a bug where the first custom labels file found was read once but any others that existed in the project were read every time a child CustomLabel was encountered.  This would result in out of memory errors when that 2nd or 3rd labels file contained thousands of labels.

https://github.com/forcedotcom/cli/issues/2939
@W-16096068@


src/convert/convertContext/recompositionFinalizer.ts
```diff
@@ -78,10 +78,27 @@ type ChildWithXml = {
 const recompose =
   (cache: XmlCache) =>
   async (stateValue: RecompositionStateValueWithParent): Promise<JsonMap> => {
-    await getXmlFromCache(cache)(stateValue.component);
+    const childComponents = stateValue.children?.toArray() ?? [];
+
+    // RecompositionState combines all labels metadata files into 1 component containing
+    // all the children.  This checks for multiple parent components and gets the xml
+    // file content from each.
+    if (stateValue.component.type.name === 'CustomLabels') {
```
mshanemc, line 86: is there a Registry attribute (strategy, recomposition etc) we could use for this conditional instead of the type name?  

Thinking of the very soon decomposition stuff for Labels that we might not want running unintentionally here

shetzel: Great point!  I think there are probably 3 things that conditional needs:
1. the component has children:  i.e., `childComponents` is truthy
2. the component recomposition strategy is `startEmpty`
3. the component recomposition transformer is `nonDecomposed`

src/convert/convertContext/recompositionFinalizer.ts
```diff
@@ -78,10 +78,31 @@ type ChildWithXml = {
...
   (cache: XmlCache) =>
   async (stateValue: RecompositionStateValueWithParent): Promise<JsonMap> => {
-    await getXmlFromCache(cache)(stateValue.component);
+    const childComponents = stateValue.children?.toArray() ?? [];
+
+    // RecompositionState combines all labels metadata files into 1 component containing
+    // all the children.  This checks for multiple parent components and gets the xml
+    // file content from each.
+    if (
+      childComponents &&
```
mshanemc, line 87: this will always be true (an empty array is still truthy).

mshanemc:
QA: 

✅  works as expected on customer repro (using npsp)

trying to break stuff
✅ empty custom labels file in apex-common
✅ all custom labels files are empty

mshanemc (approved):

PR #1369, opened 2024-07-17
@W-16265033 [MlModels] Revert some top-level types from metadata registry

neerajkumar-sfdc:
Revert metadata registries for mlmodelartifact, mlmodelconnection and mlmodelschema top level components
@W-15020853@

Functionality Before:
Able to use CLI commands for deploying/retrieving ml-model metadata in test orgs ONLY.

Functionality After:
NOT able to use CLI commands for deploying/retrieving ml-model metadata anywhere.


mshanemc (approved):
reflect metadata API coverage now

PR #1370, opened 2024-07-19
fix: handle nulls/undefined in template literal

mshanemc:
### What does this PR do?
if it's logging/debug, insert some placeholder to indicate that it's not present
for values used in "real" code, preserve existing behavior.  Many of these are from inadequate types on the metadataRegistry (ex: DEB or StaticResource have a suffix, etc)

@ W-16116290@

### DIFF
--- a/src/resolve/adapters/digitalExperienceSourceAdapter.ts
+++ b/src/resolve/adapters/digitalExperienceSourceAdapter.ts
@@ -6,6 +6,7 @@
  */
 import { dirname, join, sep } from 'node:path';
 import { Messages } from '@salesforce/core';
+import { ensureString } from '@salesforce/ts-types';
 import { META_XML_SUFFIX } from '../../common/constants';
 import { SourcePath } from '../../common/types';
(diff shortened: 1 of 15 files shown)


src/resolve/adapters/digitalExperienceSourceAdapter.ts
```diff
@@ -147,7 +148,9 @@ export class DigitalExperienceSourceAdapter extends BundleSourceAdapter {
     // 3 because we want 'digitalExperiences' directory, 'baseType' directory and 'bundleName' directory
     const basePath = pathParts.slice(0, typeFolderIndex + 3).join(sep);
     const bundleFileName = pathParts[typeFolderIndex + 2];
-    const suffix = this.isBundleType() ? this.type.suffix : this.registry.getParentType(this.type.id)?.suffix;
+    const suffix = ensureString(
```
mshanemc, line 151: this will throw, so ex: if the metadataRegistry ever loses a suffix for DEB stuff, it'll fail tests

mdonnalley (approved):

PR #1375, opened 2024-07-25
fix: manifests for custom object can omit parent

mshanemc:
### What does this PR do?

deploy -d of an entire folder including a blank CustomObject file causes the Object to be in the manifest.

Case 1: If the Object is in the manifest, the mdapi will validate that the Object metadata contains all the required fields.

Case 2: If the Object is not in the manifest (only one or more of its child types are) then the validation doesn't happen (only the child types are validated)

SDR, when generating the manifest, needs to ask if the top-level (ex: object.meta.xml) is effectively empty.  If so, that type should be omitted from the manifest (to get Case 2 behavior from the API)

### other changes
update some tests that were based on an invalid mock
update the workflow preset to make all children of workflow unaddressable

@W-16273581@ (see inv)

### DIFF
--- a/src/collections/componentSet.ts
+++ b/src/collections/componentSet.ts
@@ -406,64 +406,41 @@ export class ComponentSet extends LazyCollection<MetadataComponent> {
    * @returns Object representation of a package manifest
    */
   public async getObject(destructiveType?: DestructiveChangesType): Promise<PackageManifestObject> {
-    const version = await this.getApiVersion();
-
     // If this ComponentSet has components marked for delete, we need to
     // only include those components in a destructiveChanges.xml and
     // all other components in the regular manifest.
-    let components = this.components;
-    if (this.getTypesOfDestructiveChanges().length) {
-      components = destructiveType ? this.destructiveComponents[destructiveType] : this.manifestComponents;
-    }
+    const components = this.getTypesOfDestructiveChanges().length
+      ? destructiveType
+        ? this.destructiveComponents[destructiveType]
+        : this.manifestComponents
+      : this.components;
 
     const typeMap = new Map<string, string[]>();
 
-    const addToTypeMap = (type: MetadataType, fullName: string): void => {
-      if (type.isAddressable !== false) {
-        const typeName = type.name;
-        if (!typeMap.has(typeName)) {
-          typeMap.set(typeName, []);
-        }
-        const typeEntry = typeMap.get(typeName);
-        if (fullName === ComponentSet.WILDCARD && !type.supportsWildcardAndName && !destructiveType) {
-          // if the type doesn't support mixed wildcards and specific names, overwrite the names to be a wildcard
-          typeMap.set(typeName, [fullName]);
-        } else if (
-          typeEntry &&
-          !typeEntry.includes(fullName) &&
-          (!typeEntry.includes(ComponentSet.WILDCARD) || type.supportsWildcardAndName)
-        ) {
-          // if the type supports both wildcards and names, add them regardless
-          typeMap.get(typeName)?.push(fullName);
-        }
-      }
-    };
-
-    for (const key of components.keys()) {
+    [...components.entries()].map(([key, cmpMap]) => {
       const [typeId, fullName] = splitOnFirstDelimiter(key);
-      let type = this.registry.getTypeByName(typeId);
-
-      if (type.folderContentType) {
-        type = this.registry.getTypeByName(type.folderContentType);
-      }
-      addToTypeMap(
-        type,
-        // they're reassembled like CustomLabels.MyLabel
-        this.registry.getParentType(type.name)?.strategies?.recomposition === 'startEmpty' && fullName.includes('.')
-          ? fullName.split('.')[1]
-          : fullName
-      );
+      const type = this.registry.getTypeByName(typeId);
 
       // Add children
-      const componentMap = components.get(key);
-      if (componentMap) {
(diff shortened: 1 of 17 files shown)


src/collections/componentSet.ts
```diff
@@ -406,74 +407,60 @@ export class ComponentSet extends LazyCollection<MetadataComponent> {
...
-      addToTypeMap(
-        type,
-        // they're reassembled like CustomLabels.MyLabel
-        this.registry.getParentType(type.name)?.strategies?.recomposition === 'startEmpty' && fullName.includes('.')
-          ? fullName.split('.')[1]
-          : fullName
-      );
+      const type = this.registry.getTypeByName(typeId);
 
       // Add children
```
mshanemc, line 425: the children are always added.  the parent might or might not be depending on if it has content.

doing the children first lets the parent conditionally return early

src/collections/componentSet.ts
```diff
@@ -750,3 +737,28 @@ const splitOnFirstDelimiter = (input: string): [string, string] => {
   const indexOfSplitChar = input.indexOf(KEY_DELIMITER);
   return [input.substring(0, indexOfSplitChar), input.substring(indexOfSplitChar + 1)];
 };
+
+/** side effect: mutates the typeMap property */
+const addToTypeMap = ({
```
mshanemc, line 742: this code is simplified and now uses a set instead of an array

src/convert/convertContext/recompositionFinalizer.ts
```diff
@@ -199,19 +199,3 @@ const getXmlFromCache =
     }
     return xmlCache.get(key) ?? {};
   };
-
-/** composed function, exported from module for test */
-export const unwrapAndOmitNS =
```
mshanemc, line 204: moved for easier sharing

test/mock/type-constants/customObjectConstant.ts
```diff
@@ -34,11 +40,9 @@ export const DECOMPOSED_VIRTUAL_FS: VirtualDirectory[] = [
         name: DECOMPOSED_XML_NAME,
         data: Buffer.from(`<CustomObject xmlns="${XML_NS_URL}"><fullName>customObject__c</fullName></CustomObject>`),
       },
-      {
```
mshanemc, line 37: I don't think it should be
`objects/CustomObject__c/Fields__c.field-meta.xml`

test/mock/type-constants/customObjectConstant.ts
```diff
@@ -51,11 +55,11 @@ export const DECOMPOSED_VIRTUAL_FS: VirtualDirectory[] = [
     ],
   },
   {
-    dirPath: 'fields',
```
mshanemc, line 54: this didn't have the relative path, just the last segment

test/mock/type-constants/customObjectConstantEmptyObjectMeta.ts
```diff
@@ -0,0 +1,91 @@
+/*
```
mshanemc, line 1: copy of the existing one, but has no "real" content in the CustomObject

test/resolve/sourceComponent.test.ts
```diff
@@ -370,7 +370,9 @@ describe('SourceComponent', () => {
     );
 
     it('should return child components for a component', () => {
-      expect(decomposed.DECOMPOSED_COMPONENT.getChildren()).to.deep.equal([expectedChild, expectedChild2]);
```
mshanemc, line 373: order changed

iowillhoit:
## QA NOTES

Steps to repro
- Create an empty project
- Create a scratch org
- Open the org 
- Create a Custom Object
- Create a custom field for that object
- Retrieve with manifest
```xml
<?xml version="1.0" encoding="UTF-8"?>
<Package xmlns="http://soap.sforce.com/2006/04/metadata">
    <types>
        <members>Cat__c.Legs__c</members>
        <name>CustomField</name>
    </types>
    <version>61.0</version>
</Package>
```
- Deploy with `--source-dir force-app`
- See error `Error Cat__c Must specify a non-empty label for the CustomObject`


- 🟢 As described, seeing error before changes 
- 🟢 After changes, able to deploy with an empty parent (`Cat__c`)
- 🟢 Works with multiple different children specified (CustomField and Layout)
```xml
<types>
    <members>Cat__c-Cat Layout</members>
     <name>Layout</name>
</types>
```
- 🟡 NIT: The number of components in the status is briefly incorrect (0/2)
  - During: `Status: In Progress | ████████████████████████████████████████ | 2/2 Components`
  - After:  `Status: Succeeded | ████████████████████████████████████████ | 1/1 Components`

iowillhoit:
## QA UPDATE

- 🟢 JSON output looks good
- 🟢 Played around with this with custom source behavior and did not see any issues.
- 🟡 I am not sure if this is something we want to address, might be best to just say "Don't do this". 
  - When I retrieve a `CustomField` via a manifest, it gets the `CustomField` and the empty `CustomObject`. 
  - If I delete the empty `CustomObject`, it shows up as a `deleted` in source tracking
  - Deploying that `delete` does nothing in the org
  - If I delete the empty `CustomObject` and then reset tracking, I can still deploy changes to the `CustomField` (without its empty parent metadata file). 
  - Do we want to ignore (commit) an "empty parent" deletion in STL so that it could be removed without being tracked as a `delete`?

iowillhoit (approved):

PR #1381, opened 2024-07-30
fix: recurse through dirs then files

WillieRuemmele:
### What does this PR do?

sorts files/folders to folders first then files when searching for components so that we find everything
updates the cache when a parent with more information is found

https://github.com/forcedotcom/cli/issues/2230, @W-13629588@

### Functionality Before

would have to order modules alphabetically to get deploy to work

### Functionality After

 order doesn't matter

### DIFF
--- a/src/convert/transformers/decomposedMetadataTransformer.ts
+++ b/src/convert/transformers/decomposedMetadataTransformer.ts
@@ -49,7 +49,8 @@ export class DecomposedMetadataTransformer extends BaseMetadataTransformer {
       };
       if (component.xml && existing.component && !existing.component.xml) {
         // we've already found and created the parent of this component on L~38
-        // but now we have more information about the parent that we didn't have before, so add it
+        // but now we have more information about the parent (xml) that we didn't have before, so add it
+        // TODO: find way to determine which component has more information
         existing.component = component;
       }
       (component.getChildren() ?? []).map((child) => {
--- a/src/convert/transformers/decomposedMetadataTransformer.ts
+++ b/src/convert/transformers/decomposedMetadataTransformer.ts
@@ -47,6 +47,11 @@ export class DecomposedMetadataTransformer extends BaseMetadataTransformer {
         component,
         children: new ComponentSet([], this.registry),
       };
+      if (component.xml && existing.component && !existing.component.xml) {
+        // we've already found and created the parent of this component on L~38
+        // but now we have more information about the parent that we didn't have before, so add it
+        existing.component = component;
+      }
       (component.getChildren() ?? []).map((child) => {
         existing.children?.add(child);
       });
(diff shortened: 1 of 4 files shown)


src/convert/transformers/decomposedMetadataTransformer.ts
```diff
@@ -47,6 +47,12 @@ export class DecomposedMetadataTransformer extends BaseMetadataTransformer {
         component,
         children: new ComponentSet([], this.registry),
       };
+      if (component.xml && existing.component && !existing.component.xml) {
+        // we've already found and created the parent of this component on L~38
+        // but now we have more information about the parent (xml) that we didn't have before, so add it
+        // TODO: find way to determine which component has more information
+        existing.component = component;
+      }
```
shetzel, line 55: Is the TODO doable now or is it big enough to warrant its own WI and a separate effort?

I like the comment!

src/resolve/metadataResolver.ts
```diff
@@ -76,7 +76,20 @@ export class MetadataResolver {
...
+      // before finding the parent and returning only it - by sorting, we make it as recursive as possible
+      .sort((a, b) => {
+        if (this.tree.isDirectory(a) && this.tree.isDirectory(b)) {
+          return 0;
+        } else if (this.tree.isDirectory(a) && !this.tree.isDirectory(b)) {
+          return -1;
+        } else {
+          return 1;
+        }
+      })) {
```
shetzel, line 92: What do you think about making a sort function and referencing it here rather than all inline?  Or maybe build the array in the line above the for loop?  imho the readability decreased with everything inline.

shetzel (changes requested):

shetzel (approved):

shetzel:
QA:
The reproducer works with the fix and fails without it.  Here is what I see with the fix (SDR linked to PDR):
```
plugin-deploy-retrieve/bin/run.js project deploy start

Deploying v57.0 metadata to test-qyfpwa6ucmgs@example.com using the v62.0 SOAP API.
Deploy ID: 0AfMN0000010HMr0AM
Status: Succeeded | ████████████████████████████████████████ | 5/5 Components | Tracking: 4/4

Deployed Source
===============================================================================================================================================================
| State   Name                            Type           Path
| ─────── ─────────────────────────────── ────────────── ──────────────────────────────────────────────────────────────────────────────────────────────────────
| Created CustomObject__c.Field1__c       CustomField    force-app/main/default/objects/CustomObject__c/fields/Field1__c.field-meta.xml
| Created CustomObject__c.Field2__c       CustomField    force-app/main/default/objects/CustomObject__c/fields/Field2__c.field-meta.xml
| Created CustomObject__c                 CustomObject   force-app/main/default/objects/CustomObject__c/CustomObject__c.object-meta.xml
| Created CustomObject__c.ValidationRule2 ValidationRule force-app/aa_package/objects/CustomObject__c/validationRules/ValidationRule2.validationRule-meta.xml
| Created CustomObject__c.ValidationRule1 ValidationRule force-app/main/default/objects/CustomObject__c/validationRules/ValidationRule1.validationRule-meta.xml
```

PR #1383, opened 2024-08-02
Wr/sccache

WillieRuemmele:
@W-16383577@

keeps a map of path=>content on the source component to avoid re-reading any files and referencing information we've already gathered.


cristiand391 (approved):

cristiand391:
No manual QA, NUTs and UTs should be enough for this.

PR #1392, opened 2024-08-09
feat: custom label beta 2

mshanemc:
### What does this PR do?

after beta feedback https://github.com/forcedotcom/cli/issues/2840 and discussion https://github.com/forcedotcom/cli/discussions/2948, update the way labels decompose.

Meta
- `CustomLabels` and `CustomLabel` both use the default Adapter (no change) and can stand alone as SourceComponents.  There is no relationship between them, except for the output of the Transformers
- `CustomLabel` is always used for source format (1 file per label)
- `CustomLabels` is always uses for metadata format (1 file with all the labels in it.  named `CustomLabels`)
- since the mergeSet will be a bunch of customLabel, but the retrieve is CustomLabels, `toSourceFormat` needs the entire mergeSet (and not a single `mergeWith` SourceComponent) to find components with the matching fullName (for labels outside the default dir, to avoid creating duplicates).

change the decomposeCustomLabelsBeta to do one file per label (preset update)
- 2 new transformer classes in file to handle label/labels
- new finalizer for labels
- the toSourceFormat method now takes an object (component, mergeComponent, and [new] the entire mergeSet)
- presets can put an empty string `""` for suffix/childTypes/strictDirs to **remove** an entry from the default registry (this preset removes the parent/child relationship between label/labels)

other
- TODOs for removing exported Transformers in the next major.  These aren't used that I could see and should never have been exported.

## QA notes
there are some small changes in STL necessary to make this behave right https://github.com/forcedotcom/source-tracking/pull/651 and you'll need those for testing tracking.

### Tested during development

- local create, preview, deploy
- deploy via mpd
- remote update, track, retrieve
- remote delete, track, retrieve
- local delete, track, retrieve
- retrieve into mpd (existing local labels with remote change update, new labels are placed in default dir, can do both)
- label file outside of the `/labels` folder in source format (deploy, retrieve using -m CustomLabel:foo)
- subfolders within the `/labels` directory for grouping labels (deploy, retrieve changes) 
- retrieve using `--metadata CustomLabels` ("give me all of them") doesn't match the local CustomLabel and put them in the right places (they'll all end up in the default directory).  Workaround `--metadata CustomLabel` (using `--metadata CustomLabels` now throws an error when the new Transformer is used).

@W-16375268@


src/convert/convertContext/decomposedCustomLabelFinalizer.ts
```diff
@@ -0,0 +1,76 @@
...
+  customLabelByFullName: Map<string, CustomLabel>;
+};
+
+/**
+ * Merges child components that share the same parent in the conversion pipeline
+ * into a single file.
+ *
+ * Inserts unclaimed child components into the parent that belongs to the default package
+ */
+export class DecomposedCustomLabelsFinalizer extends ConvertTransactionFinalizer<CustomLabelState> {
```
shetzel, line 30: The class name doesn't match the file name.

src/convert/transformers/metadataTransformerFactory.ts
```diff
@@ -32,15 +32,19 @@ export class MetadataTransformerFactory {
     const type = component.parent ? component.parent.type : component.type;
     const transformerId = type.strategies?.transformer;
     switch (transformerId) {
-      case TransformerStrategy.Standard:
+      case 'standard':
```
shetzel, line 35: why don't you like the `TransformerStrategy.Standard` style reference?  Seems easier for refactors and having 1 place to look for the possible transformers.

src/index.ts
```diff
@@ -90,15 +90,10 @@ export {
...
-  MetadataRegistry,
-  MetadataType,
-  DecompositionStrategy,
-  RecompositionStrategy,
-  TransformerStrategy,
-} from './registry';
+
+export { RegistryAccess, registry, getCurrentApiVersion, MetadataRegistry, MetadataType } from './registry';
+
+// TODO: don't export these strategies
```
shetzel, line 96: Is there a work item for this?

src/registry/presets/decomposeCustomLabelsBeta.json
```diff
@@ -1,44 +1,32 @@
 {
   "childTypes": {
-    "customlabel": "customlabels"
+    "customlabel": ""
```
shetzel, line 3: This feels like hidden magic.  Is it explained what this means anywhere?

shetzel (approved):

src/convert/transformers/metadataTransformerFactory.ts
mshanemc: TS type gives you autocomplete and validates that it's one of the possible values.  If you wanted to change the name everywhere, you'd just change the type.

src/index.ts
mshanemc: No.  It's more of a thing I look around for when I do a major of a library.

src/registry/presets/decomposeCustomLabelsBeta.json
mshanemc: https://github.com/forcedotcom/source-deploy-retrieve/pull/1392/files#diff-61c4540d4561b7af0f31a1721a2f90d455a9f7f270675fc9feaa024b6ba022a7R175

mshanemc: [yeah, I sometimes wish json could have comments!]

src/convert/transformers/metadataTransformerFactory.ts
mshanemc: I don't really like enums in TS

src/convert/convertContext/decomposedCustomLabelFinalizer.ts
mshanemc: nice catch, I'll fix that

PR #1452, opened 2024-11-01
fix(resolver): ignore non-component XML files in project

cristiand391:
### What does this PR do?

Updates the metadata resolver to ignore XML files in a project that aren't part of a source component.

A customer had datapacks with an XML file that isn't part of the metadata inside `force_app`, when SDR built the component to deploy a specific metadata it would try to resolve the XML file as `emailservicesfunction` because the helpers to detect if a filepath was a metadata type were checking if the file suffix matched one in the registry (`emailservicesfunction.suffix === 'xml'`).

This PR updates `isMetadata` to first try to find the type based on the suffix, if it there's not then it will check for `strictDirectoryNames` type ones **and** their `directoryName`.

Repro:

1. create an invalid file in the project dir (force-app/main/default/lala/hehe.xml)
2. update forceignore to ignore the file (**/lala/**)
3. confirm it shows up when running `sf project list ignored`
4. try deploying something via --metadata: `sf project deploy start -m ApexClass` 
5. see error about SDR trying to resolve an it as `emailservicesfunction`
6. rename the file extension from file created in step one to something else other than `xml`, then step 4 works b/c it doesn't try to resolve it as a component.

@W-17113942@

### DIFF
--- a/src/resolve/metadataResolver.ts
+++ b/src/resolve/metadataResolver.ts
@@ -373,7 +373,16 @@ const parseAsContentMetadataXml =
     const suffixType = registry.getTypeBySuffix(extName(fsPath));
     if (!suffixType) return false;
 
-    return fsPath.split(sep).includes(suffixType.directoryName);
+    const matchesSuffixType = fsPath.split(sep).includes(suffixType.directoryName);
+    if (matchesSuffixType) return matchesSuffixType;
+
+    // it might be a type that requires strict parent folder name.
+    const strictFolderSuffixType = registry
+      .getStrictFolderTypes()
+      .find((l) => l.suffix === suffixType.suffix && l.directoryName && l.name !== suffixType.name);
+    if (!strictFolderSuffixType) return false;
+
+    return fsPath.split(sep).includes(strictFolderSuffixType.directoryName);
   };
 
 /**
--- a/src/resolve/metadataResolver.ts
+++ b/src/resolve/metadataResolver.ts
@@ -369,8 +369,12 @@ const resolveType =
  */
 const parseAsContentMetadataXml =
   (registry: RegistryAccess) =>
-  (fsPath: string): boolean =>
-    Boolean(registry.getTypeBySuffix(extName(fsPath)));
+  (fsPath: string): boolean => {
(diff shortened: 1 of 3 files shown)


src/resolve/metadataResolver.ts
```diff
@@ -369,8 +369,21 @@ const resolveType =
...
 const parseAsContentMetadataXml =
   (registry: RegistryAccess) =>
-  (fsPath: string): boolean =>
-    Boolean(registry.getTypeBySuffix(extName(fsPath)));
+  (fsPath: string): boolean => {
+    const suffixType = registry.getTypeBySuffix(extName(fsPath));
+    if (!suffixType) return false;
+
+    const matchesSuffixType = fsPath.split(sep).includes(suffixType.directoryName);
+    if (matchesSuffixType) return matchesSuffixType;
```
cristiand391, line 377: for an XML file, `suffixType` would return the `emailservicesfunction` entry:
https://github.com/forcedotcom/source-deploy-retrieve/blob/9d94840fa9523c2e9d4268f231c7f889ab8160b0/src/registry/metadataRegistry.json#L522

src/resolve/metadataResolver.ts
```diff
@@ -369,8 +369,21 @@ const resolveType =
...
+
+    // it might be a type that requires strict parent folder name.
+    const strictFolderSuffixType = registry
+      .getStrictFolderTypes()
+      .find((l) => l.suffix === suffixType.suffix && l.directoryName && l.name !== suffixType.name);
```
cristiand391, line 382: if `registry.getTypeBySuffix` entry didn't match, check for strictFolderTypes.

snapshot tests caught an issue with `CustomSite` where `getTypeBySuffix` would return `sitedotcom` (no match with the `directoryName`), this will return the strict folder types -> find if one with directoryName and suffix matches (and isn't `suffixType`).
https://github.com/forcedotcom/source-deploy-retrieve/blob/9d94840fa9523c2e9d4268f231c7f889ab8160b0/src/registry/metadataRegistry.json#L459

shetzel (approved):

shetzel:
Manual tests with the changes passed.  I added a UT that fails without the changes so if CI is green we can merge.

PR #1467, opened 2024-12-13
fix: add workflow flow actions to decomposed workflow preset and allow isAddressable

mcarvin8:
### What does this PR do?

Updates the decomposed workflow preset to add the child type: Workflow Flow Actions

The overall metadata registry is also missing this child type so I'm adding it to that JSON as well.

In the current decomposed preset, Workflow Flow Actions are added to the `*.workflow-meta.xml` file instead of decomposed into their own directory and smaller files.

This is causing issues deploying decomposed workflow children, as Salesforce CLI appears to check the `*.workflow-meta.xml` file even when deploying just 1 decomposed workflow child like an email alert. Since the  `*.workflow-meta.xml`  isn't "empty", I'm unable to deploy just a decomposed WorkflowAlert for example.

Also, the "isAddressable: false" key in each supported decomposed workflow child is preventing each decomposed workflow child from being retrieved or deployed without the parent. 

After working with @shetzel via CLI Issue 2563, updating the preset to remove all `isAddressable: false` lines and manually add workflow flow actions to the preset fixes my issue retrieving and deploying decomposed workflow children. I've tested this out in my work's repo, which has tons of workflows with workflow flow actions.

Let me know if your team wants to make this a `beta2` instead of modfying an existing `beta`. Since I'm not changing the overall decomposed workflow structure, just adding a missing child and fixing the addressable feature, I think it makes sense to me to just update the `beta` preset.

https://github.com/forcedotcom/cli/issues/2563

### Functionality Before

Decomposed workflow children besides Flow Actions are decomposed into their own sub-directories and files.

Flow actions are added directly to the `*.workflow-meta.xml` files.

Retrieves can work but deployments fail because it's seeing the Flow Action in the `*.workflow-meta.xml` file and `isAddressable` is false, so I can't deploy the child without the parent file.

### Functionality After

All decomposed workflow children including Flow Actions are decomposed into their own sub-directories and files.

The `*.workflow-meta.xml` file just contains the workflow header/footer.

Retrieving and deploying individual workflow children including Flow Actions works with Salesforce CLI.

[skip-validate-pr]

### DIFF
--- a/src/registry/metadataRegistry.json
+++ b/src/registry/metadataRegistry.json
@@ -29,6 +29,7 @@
     "weblink": "customobject",
     "workflowalert": "workflow",
     "workflowfieldupdate": "workflow",
+    "workflowflowaction": "workflow",
     "workflowknowledgepublish": "workflow",
     "workflowoutboundmessage": "workflow",
     "workflowrule": "workflow",
@@ -513,6 +514,7 @@
     "workflow": "workflow",
     "workflowAlert": "workflowalert",
     "workflowFieldUpdate": "workflowfieldupdate",
+    "workflowFlowAction": "workflowflowaction",
     "workflowKnowledgePublish": "workflowknowledgepublish",
     "workflowOutboundMessage": "workflowoutboundmessage",
     "workflowRule": "workflowrule",
@@ -4517,6 +4519,7 @@
         "directories": {
           "workflowAlerts": "workflowalert",
           "workflowFieldUpdates": "workflowfieldupdate",
+          "workflowFlowActions": "workflowflowaction",
           "workflowKnowledgePublishs": "workflowknowledgepublish",
           "workflowOutboundMessages": "workflowoutboundmessage",
           "workflowRules": "workflowrule",
@@ -4526,6 +4529,7 @@
         "suffixes": {
           "workflowAlert": "workflowalert",
           "workflowFieldUpdate": "workflowfieldupdate",
+          "workflowFlowAction": "workflowflowaction",
           "workflowKnowledgePublish": "workflowknowledgepublish",
           "workflowOutboundMessage": "workflowoutboundmessage",
           "workflowRule": "workflowrule",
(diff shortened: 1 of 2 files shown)


BluecanvasReen:
By coincidence I have been debugging this problem for the last couple of days. The same problem happens for SharingRules.

The normal registry, does not correctly detect the fullname of the components, because it does not contain the correct strategy, and the beta version does not work because of isAddressable === false.

Added this comment to get notifications, but happy to support this issue if need be.

mcarvin8:
You can work around the `isAddressable` issue @BluecanvasReen if you copy the existing [preset ](https://github.com/forcedotcom/source-deploy-retrieve/blob/main/src/registry/presets/decomposeSharingRulesBeta.json) into your `sfdx-project.json` file in a key named `registryCustomizations` and remove all `isAddressable === false` lines.

I have my `sfdx-project.json` with the updated workflow preset in https://github.com/forcedotcom/cli/issues/2563 you can reference.

Obviously the long-term fix is for Salesforce to fix the beta preset with a new CLI build, but doing this with workflows was able to get me un-blocked in my work's repo.

mcarvin8:
@shetzel - Can this request be reviewed soon please by you or someone else on the CLI team?

shetzel:
> @shetzel - Can this request be reviewed soon please by you or someone else on the CLI team?

Hi Matt - sorry for the delay.  Thanks for making the PR.  I'll get someone to look.  We've been in a release moratorium.

shetzel (approved):

WillieRuemmele:
Hi @mcarvin8 - I've started running our tests on your fork over [here](https://github.com/forcedotcom/source-deploy-retrieve/pull/1476) - I had to update our snapshots because your changes affected those, you can run those `yarn test:snapshot` to see what they're now producing with you registry+preset changes, let's make sure those look correct

mcarvin8:
Thank you @shetzel and @WillieRuemmele ! Yea I just realized today that I should update the overall registry JSON as well for the workflow flow actions since it's missing there as well.

When I run `yarn test:snapshot` on my branch, the changes look good to me.

mcarvin8:
@BluecanvasReen - Sorry your issue with SharingRules got pushed to the side in all of this.

I'm assuming Salesforce would probably advocate for an issue in their issues repo to update the preset

Is this update just removing the `isAddressable` lines in the decompose preset for sharing rules? Or is there updates needed to the overall metadata registry?

I could push the change for you in my forked SDR repo, but the Salesforce CLI team might just want a new issues created in https://github.com/forcedotcom/cli/issues

shetzel:
@BluecanvasReen @mcarvin8 - please review when you have a sec.  https://github.com/forcedotcom/source-deploy-retrieve/pull/1482

Thanks Matt for bringing this up.  I completely missed it!  Although we'd prefer a new issue, this will work too.

PR #1482, opened 2025-01-09
fix: sharing rule children are now addressable in the beta preset

shetzel:
### What does this PR do?
sharing rule children are now addressable in the beta preset

@W-17572243@

### DIFF
--- a/src/registry/presets/decomposeSharingRulesBeta.json
+++ b/src/registry/presets/decomposeSharingRulesBeta.json
@@ -30,7 +30,6 @@
           "sharingcriteriarule": {
             "directoryName": "sharingCriteriaRules",
             "id": "sharingcriteriarule",
-            "isAddressable": false,
             "name": "SharingCriteriaRule",
             "suffix": "sharingCriteriaRule",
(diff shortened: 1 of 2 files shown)


WillieRuemmele (approved):

mcarvin8:
Looks good to me @shetzel !

BluecanvasReen:
@WillieRuemmele apologies for the late reply, have been OOO for a holiday. Nice progress on the fixes.

Seems like something funky is going on for me. If I convert Object + SharingCriteriaRule to source format (aka pre-deploy), it seems like the properties of the SharingCriteriaRule gets added to the Object.

I think its because of the finalizer, that has a map per name, not type + name:
https://github.com/forcedotcom/source-deploy-retrieve/blob/main/src/convert/transformers/decomposedMetadataTransformer.ts#L38
After attaching a debugger I've found that "transactionState" contains both SharingRule & Object child entities.

If I just deploy the SharingRules or just deploy the Object, everything works.

BluecanvasReen:
Oops wrong mention @mcarvin8 @shetzel

PR #1496, opened 2025-01-31
fix: properly merge `DigitalExperience` components W-16646427

cristiand391:
### What does this PR do?

Fixes SDR retrieving and writing `DigitalExperience` metadata to the wrong paths in this scenario:
1. MPD project with 1 DE in the default pkg and another one in a non-default pkg dir (all defined in `sfdx-project.json`)
2. retrieving the specific DE in the non-default dir or its DEB parent.

Given this project structure:
```
➜  mpdProject git:(cd/deb-mpd-nut) ✗ tree -L 6 -P digitalExperiences
.
├── config
├── force-app
│   └── main
│       └── default
│           ├── classes
│           ├── digitalExperienceConfigs
│           ├── digitalExperiences
│           │   └── site
│           │       └── Capricorn_Coffee_A1
│           ├── networks
│           ├── pages
│           └── sites
├── my-app
│   └── main
│       └── default
│           └── digitalExperiences
│               └── site
│                   └── Capricorn_Coffee_B1
```

where `force-app` (default pkg dir) and `my-app` contain 1 DE each, if you retrieve one of `Capricorn_Coffee_B1`'s files  `sf`/SDR writes the retrieve to `force-app` instead of merging it with the contents in `my-app`:

Screenshot 2025-02-06 at 12 53 55

### Bug

`project retrieve start` builds the correct source-backed DE components for merging here:
https://github.com/salesforcecli/plugin-deploy-retrieve/blob/0bc486a28e3d4b49f42093144b3e435d34404ffe/src/commands/project/retrieve/start.ts#L171

but the metadata converter prefers the component parent if present for merging (I assume for decomposed types that aren't individually addressable):
https://github.com/forcedotcom/source-deploy-retrieve/blob/68c3a4e034ee0dbb09c33df974d47e0ff78769d2/src/convert/metadataConverter.ts#L183C18-L183C34

so the metadata transformer would always try to merge a retrieved DE against its local DEB (no match so the retrieved DE would get saved to the default pkg dir):
https://github.com/forcedotcom/source-deploy-retrieve/blob/68c3a4e034ee0dbb09c33df974d47e0ff78769d2/src/convert/streams.ts#L76

### Fix
* Make the metadata converter prefer a DE component over its parent when building the merge set.
* Update the default transformer to properly calculate the source destination when merging DEs

### Testing
This PR adds a small UT for the converter fix and https://github.com/salesforcecli/plugin-deploy-retrieve/pull/1279 includes a full MPD project with NUTs to test component merging of DEB/single DEs retrieves.

1. checkout https://github.com/salesforcecli/plugin-deploy-retrieve/pull/1279 
2. `cd test/nuts/digitalExperienceBundle/mpdProject`
3. create org: `sf org create scratch -d -f config/project-scratch-def.json -a deb-mpd`
4. full deploy: `sf project deploy start`
5. retrieve the DE in `my-app`, see `sf` writes it into `force-app`

@W-16646427@

### DIFF
--- a/src/convert/metadataConverter.ts
+++ b/src/convert/metadataConverter.ts
@@ -180,7 +180,11 @@ function getMergeConfigOutputs(
   const mergeSet = new ComponentSet(undefined, registry);
   // since child components are composed in metadata format, we need to merge using the parent
   for (const component of output.mergeWith) {
-    mergeSet.add(component.parent ?? component);
+    if (component.type.id === 'digitalexperience' && !component.parent?.content) {
+      mergeSet.add(component);
+    } else {
+      mergeSet.add(component.parent ?? component);
+    }
   }
   const writer = new StandardWriter(output.defaultDirectory);
 
--- a/src/convert/metadataConverter.ts
+++ b/src/convert/metadataConverter.ts
@@ -178,11 +178,12 @@ function getMergeConfigOutputs(
   }
   const defaultDirectory = output.defaultDirectory;
   const mergeSet = new ComponentSet(undefined, registry);
-  // since child components are composed in metadata format, we need to merge using the parent
   for (const component of output.mergeWith) {
     if (component.type.id === 'digitalexperience' && !component.parent?.content) {
+      // DE is addressable without its parent (DEB)
       mergeSet.add(component);
     } else {
+      // since child components are composed in metadata format, we need to merge using the parent
       mergeSet.add(component.parent ?? component);
     }
   }
--- a/src/convert/metadataConverter.ts
+++ b/src/convert/metadataConverter.ts
@@ -179,7 +179,7 @@ function getMergeConfigOutputs(
   const defaultDirectory = output.defaultDirectory;
   const mergeSet = new ComponentSet(undefined, registry);
   for (const component of output.mergeWith) {
-    if (component.type.id === 'digitalexperience' && !component.parent?.content) {
+    if (component.type.strategies?.adapter === 'digitalExperience' && !component.parent?.content) {
       // DE is addressable without its parent (DEB)
       mergeSet.add(component);
     } else {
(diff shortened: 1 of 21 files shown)


Princejoroe (approved):

src/convert/metadataConverter.ts
```diff
@@ -178,9 +178,14 @@ function getMergeConfigOutputs(
   }
   const defaultDirectory = output.defaultDirectory;
   const mergeSet = new ComponentSet(undefined, registry);
-  // since child components are composed in metadata format, we need to merge using the parent
   for (const component of output.mergeWith) {
-    mergeSet.add(component.parent ?? component);
+    if (component.type.id === 'digitalexperience' && !component.parent?.content) {
```
WillieRuemmele, line 182: does this only happen for DEs? are there other bundle types? you could actually check `component.type.strategies.adapter==='digitalExperience'` so it would apply to all of the types that use that adapter, which I think is only DEB anyways

src/convert/transformers/defaultMetadataTransformer.ts
```diff
@@ -73,8 +73,13 @@ const getContentSourceDestination = (
 ): SourcePath => {
   if (mergeWith?.content) {
     if (component.content && component.tree.isDirectory(component.content)) {
-      const relative = trimUntil(source, basename(component.content));
-      return join(dirname(mergeWith.content), relative);
+      // DEs are always inside a dir.
+      if (component.type.id === 'digitalexperience') {
```
WillieRuemmele, line 77: same idea as above here

WillieRuemmele (approved):

src/convert/metadataConverter.ts
cristiand391: > does this only happen for DEs?
technically it should happen with any child type that's addressable without its parent, but yes only DE/DEB use this adapter.
In the investigation the customer mentioned `sitedotcom` sites were working ok on MPDs, those don't have childs.

> you could actually check component.type.strategies.adapter==='digitalExperience'
done ✅

WillieRuemmele (approved):

PR #1518, opened 2025-03-03
fix(@W-17943542): fix content path for inline media file inside nested folders

skumar846676:
### What does this PR do?

LWC content can have inline media files in nested folder, updated the adapter to get correct content path for such files.

@https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00002AYQG4YAP/view@

### Functionality Before

DigitalExperience did not have any files in nested folder.

### Functionality After

DigitalExperience have files in nested folder for inline media content types.


src/resolve/adapters/digitalExperienceSourceAdapter.ts
```diff
@@ -164,6 +166,5 @@ export class DigitalExperienceSourceAdapter extends BundleSourceAdapter {
  * @returns name of type/apiName format
  */
 const calculateNameFromPath = (contentPath: string): string => `${parentName(contentPath)}/${baseName(contentPath)}`;
-
-// Bundle hierarchy baseType/spaceApiName/contentType/contentApiName/variantFolders/file
-const digitalExperienceBundleWithVariantsDepth = 5;
+const digitalExperienceStructure = 'BaseType/SpaceApiName/ContentType/ContentApiName';
+const contentParts = digitalExperienceStructure.split(sep);
```
shetzel, line 170: `sep` is different depending on the OS, but you're splitting on a posix path.  This will not work on Windows systems.  You either have to use `join` on the individual paths of `digitalExperienceStructure` or not use `sep` and instead do something like `.split('/')`

skumar846676: Thanks!! updated digitalExperienceStructure.

PR #1535, opened 2025-04-04
W-17980476: exclude folder when folder content is excluded from org connection

EstebanRomero84:
### What does this PR do?

Updates CSB build() method for when an Org and excluded types are passed in. It will check if the excluded type is a folder content and if so, it will also exclude the folder from the resulting CS. 
E.g. if Report type is excluded, it will also exclude ReportFolder.

[#3232](https://github.com/forcedotcom/cli/issues/3232), [@W-17980476@ >](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00002AjRNfYAN/view)

### DIFF
--- a/test/collections/componentSetBuilder.test.ts
+++ b/test/collections/componentSetBuilder.test.ts
@@ -623,6 +623,63 @@ describe('ComponentSetBuilder', () => {
         expect(compSet.has(apexClassWildcardMatch)).to.equal(true);
       });
 
+      it('should exclude folder type when excluding a type that is stored in folders', async () => {
+        const packageDir1 = path.resolve('force-app');
+        const registry = new RegistryAccess();
+        const dashboardType = registry.getTypeByName('Dashboard');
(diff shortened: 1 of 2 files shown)


test/collections/componentSetBuilder.test.ts
```diff
@@ -623,6 +623,63 @@ describe('ComponentSetBuilder', () => {
         expect(compSet.has(apexClassWildcardMatch)).to.equal(true);
       });
 
+      it('should exclude folder type when excluding a type that is stored in folders', async () => {
```
WillieRuemmele, line 626: can we add a test that excludes `DasbhoardFolder` and makes sure `Dashboard` is still included?

WillieRuemmele:
### QA Notes
---
✅ : running `project generate manifest --from-org test-1iuewepdutjr@example.com --excluded-metadata Report` created a manifest without a `Report` entry, before, reports were present

PR #1571, opened 2025-06-03
fix: change map to use name#type for key  @W-18476014@

WillieRuemmele:
### What does this PR do?

in the recomposition finalizer map, it was setting the components based on name, not name and type

changed the map to use name+type to avoid merging different types with the same name

https://github.com/forcedotcom/cli/issues/3275, @W-18476014@

### Functionality Before

could no deploy decomposed workflows and custom objects with the same name

### Functionality After

can deploy that


soridalac (approved):

soridalac:
**QA notes**:
✅ : Deployed source successfully on decomposed workflows together with CustomObjects.

PR #1597, opened 2025-08-01
W-19183524 fix: web bundleability

mshanemc:
### What does this PR do?
Make SDR more web-friendly

1. remove the isBinaryFile dependency (was causing bundling problems) and simplify to only what we need
2. defer `promisify` of pipeline by wrapping it in a fn (another bundle issue with streams polyfills)
3. forceIgnore's upward search caused an infinite loop on the web (never finds the `/`).  Limit it to 25 levels of folders 

see other comments for addiontal notes

@W-19183524@


src/utils/fileSystemHandler.ts
```diff
@@ -9,16 +9,8 @@ import * as path from 'node:path';
 import fs from 'graceful-fs';
 import { SourcePath } from '../common/types';
```
mshanemc, line 11: benefit of the doubt: probably written before recursive mkdir was available

PR #1604, opened 2025-08-16
W-19263098 fix: web perf

mshanemc:
### What does this PR do?
make web retrieve faster, see comments

@W-19263098@

### DIFF
--- a/test/utils/fileSystemHandler.test.ts
+++ b/test/utils/fileSystemHandler.test.ts
@@ -8,7 +8,7 @@ import { join } from 'node:path';
 import { SinonStub, createSandbox } from 'sinon';
 import { expect, config } from 'chai';
 import fs from 'graceful-fs';
-import { ensureFileExists, searchUp } from '../../src/utils/fileSystemHandler';
+import { searchUp } from '../../src/utils/fileSystemHandler';
 
 const env = createSandbox();
 config.truncateThreshold = 0;


src/client/deployMessages.ts
```diff
@@ -6,7 +6,7 @@
  */
 
 import { basename, dirname, extname, join, posix, sep } from 'node:path/posix';
-import { SfError } from '@salesforce/core';
+import { SfError } from '@salesforce/core/sfError';
```
mshanemc, line 9: there a bunch of noise related to these top-level imports.  I'll comment on the other interesting bits.

src/convert/streams.ts
```diff
@@ -159,7 +160,7 @@ export class StandardWriter extends ComponentWriter {
           chunk.writeInfos
             .map(makeWriteInfoAbsolute(this.rootDestination))
             .filter(existsOrDoesntMatchIgnored(this.forceignore, this.logger)) // Skip files matched by default ignore
-            .map((info) => {
+            .map(async (info) => {
```
mshanemc, line 163: async because `ensurefileExists` is now async

src/utils/fileSystemHandler.ts
```diff
@@ -9,10 +9,9 @@ import * as path from 'node:path';
 import fs from 'graceful-fs';
 import { SourcePath } from '../common/types';
 
-export function ensureFileExists(filePath: string): void {
-  fs.mkdirSync(path.dirname(filePath), { recursive: true });
-  fs.closeSync(fs.openSync(filePath, 'w'));
-}
+export const ensureFileExists = async (filePath: string): Promise<void> => {
+  await fs.promises.mkdir(path.dirname(filePath), { recursive: true });
```
mshanemc, line 13: creating writeStreams will create a file.  the mkdir is still necessary to make sure it has a folder to be created in, but the open/close is pointless

PR #1699, opened 2026-03-05
fix: replacements splitted between chunks W-20635868

Bartheleway:
### What does this PR do?

When using replacement strings on big files, it happen a replacement string is split in two different chunks thus not being replaced.

The modification consider treating only complete lines as replacement string should not be split on several lines. Remaining characters are added to the next chunk in order to be processed as well.

https://github.com/forcedotcom/cli/issues/3461, @W-20635868@

### Functionality Before

Replacement process miss some tokens.

### Functionality After

All tokens are correctly replaced.
Tested locally on the file I had issue with and it solved it without causing other impact.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "node": ">=18.0.0"
   },
   "dependencies": {
-    "@salesforce/core": "^8.31.0",
+    "@salesforce/core": "^8.31.2",
     "@salesforce/kit": "^3.2.4",
     "@salesforce/ts-types": "^2.0.12",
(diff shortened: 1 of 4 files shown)


WillieRuemmele:
Hi @Bartheleway - thanks for the contribution! I'm adding a few tests to verify the behavior, but will merge this PR! and keep an eye on the release notes for a shoutout!

WillieRuemmele:
CI all passed here : https://github.com/forcedotcom/source-deploy-retrieve/pull/1793

PR #1794, opened 2026-06-25
fix: improve polling frequency calculation @W-21321719@ @W-20866522@

WillieRuemmele:
## Summary

- Use deploy zip size (bytes) to determine polling frequency for large deploys (>1MB), preventing aggressive polling on deploys that take minutes to complete
- Raise the minimum polling frequency from 100ms to 1000ms for all deploys — the sub-second intervals (100/250/500ms) provide no practical benefit given typical metadata API response times
- Small deploys are unaffected (1s floor is still fast enough for sub-second server-side completions)

## Polling Frequency Tiers

| Condition | Frequency |
|-----------|-----------|
| Zip > 15MB | 10s |
| Zip 5–15MB | 5s |
| Zip 1–5MB | 2s |
| ≤ 1MB zip or no zip, ≤ 1000 components | 1s |
| ≤ 1MB zip or no zip, > 1000 components | scales with component count |

## Telemetry Validation

Analyzed 30 days of `metadata_api_deploy_result` telemetry data. The vast majority of deploys are small (median zip ~7KB) and were polling at an average of ~200ms — well below useful thresholds. The new 1s floor significantly reduces unnecessary API calls for the most common case without impacting perceived latency. Large deploys (>15MB) were polling at ~6-7s via component count scaling; the explicit 10s tier better matches their actual completion time.

## Work Items

- @W-21321719@: Reduce checkDeployStatus polling frequency — large deploys were polling at 100ms
- @W-20866522@: Simplify `calculatePollingFrequency` with 1000ms minimum floor

## Test plan

- [x] All existing polling frequency tests updated and passing
- [x] New tests for zip-size-based frequency tiers and boundary conditions
- [x] Type check clean
- [x] Lint clean (warnings only, all pre-existing)
- [ ] Verify no regression in deploy UX for small deploys (1s poll is imperceptible)
- [ ] Verify large deploy polling is noticeably less aggressive

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "node": ">=18.0.0"
   },
   "dependencies": {
-    "@salesforce/core": "^8.31.2",
+    "@salesforce/core": "^8.31.4",
     "@salesforce/kit": "^3.2.4",
     "@salesforce/ts-types": "^2.0.12",
     "@salesforce/types": "^1.6.0",
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "yaml": "^2.9.0"
   },
   "devDependencies": {
-    "@jsforce/jsforce-node": "^3.10.15",
+    "@jsforce/jsforce-node": "^3.10.16",
     "@salesforce/cli-plugins-testkit": "^5.3.58",
     "@salesforce/dev-scripts": "^11.0.4",
     "@types/deep-equal-in-any-order": "^1.0.1",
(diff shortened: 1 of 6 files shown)


soridalac (approved):

soridalac:
**QA notes**:
Create small-test and large-test:

✅ : Minimum 1000ms polling:
componentCount: 1, zipSize: 841 bytes and returning 1000ms (componentCount <= 1000)
✅ :  Zip-size based polling (>15MB = 10s)
componentCount: 5, zipSize: 20980677 bytes (~21MB) and returning 10000ms for large deploy (zipSize > 15MB)

PR #1795, opened 2026-06-26
Wr/fix heap oom large custom labels @W-21251439@

WillieRuemmele:

#<Insert GitHub Issue>, @W-21251439@

### Functionality Before

<insert gif and/or summary>

### Functionality After

<insert gif and/or summary>

### Metadata Registry: Successful Deploy and Retrieve (required if applicable)

**Deploy output:**

<paste CLI output and/or screenshot>

**Retrieve output:**

<paste CLI output and/or screenshot>

---

### Problem

When multiple package directories contain the same `CustomLabels` file (common in multi-package projects), the recomposition finalizer (source→metadata conversion for deploy) was accumulating duplicate children. With large label files across 2-3 packages, this caused the merged output to double or triple in size, leading to heap OOM crashes.

### Fix: `recompositionFinalizer.ts`

Added `deduplicateChildren()` — when the same label (keyed by `type.name#fullName`) appears from multiple packages, only the first occurrence is kept. This is safe because:

- The merged `CustomLabels.labels` is an ephemeral artifact for deploy (zipped and sent to the org, never written to disk).
- Retrieve uses `nonDecompositionFinalizer`, which writes labels back to their respective package directories — users never see git diffs.
- Identical labels in multiple packages are the same logical metadata member; deploying it twice was already a no-op on the org side.

### Fix: `nonDecompositionFinalizer.ts`

Refactored `initMergeMap` to parse each parent XML file once and extract all children directly, rather than re-parsing the parent for every individual child component. This eliminates redundant XML parsing that scaled quadratically with label count.

### Snapshot diff (~16k lines)

The `customLabels-multiple` test snapshot went from 32,918 lines to 16,471 — almost exactly halved. The test data has labels across 2 packages (`second-app` and `third-app`) with largely identical content, so the old behavior was duplicating nearly every entry. The ~92 "additions" in the diff are ordering artifacts from the dedup changing sort positions, not new labels.

### New test: `customLabels-merged`

A focused snapshot test with small synthetic data (3 packages, 8 unique labels, 2 shared across packages) to verify the dedup behavior explicitly without relying on the large real-world test data.


src/convert/convertContext/nonDecompositionFinalizer.ts
```diff
@@ -159,37 +160,32 @@ export class NonDecompositionFinalizer extends ConvertTransactionFinalizer<NonDe
...
-          await getMappedChildren(c),
-        ]
-      )
-    );
-
-    this.mergeMap = new Map(result);
+    for (const component of allComponentsOfType) {
+      const xmlPath = ensureString(component.xml, `Missing xml file for ${component.type.name}`);
+      // eslint-disable-next-line no-await-in-loop
+      const parentXml = await component.parseXml();
```
jfeingold35: If you want to get rid of this `eslint-disable` statement, you could do something like
```
parsedComponents = await Promise.all(allComponentsOfType.map(c => c.parseXml()));
for (const parsedComponent of parsedComponents) {
...
}
```

src/convert/convertContext/recompositionFinalizer.ts
```diff
@@ -109,15 +109,15 @@ const recompose =
...
-          xmlContents: await getXmlFromCache(cache)(child),
-          groupName: getXmlElement(child.type),
-        })
-      )
-    );
+    const childXmls: ChildWithXml[] = [];
+    for (const child of childComponents.filter(ensureMetadataComponentWithParent)) {
+      childXmls.push({
+        cmp: child,
+        // eslint-disable-next-line no-await-in-loop
```
jfeingold35, line 116: Was there a way to fix the bug that didn't involve removing Promise.all?

src/convert/streams.ts
```diff
@@ -286,6 +293,7 @@ export class JsToXml extends Readable {
       commentPropName: XML_COMMENT_PROP_NAME,
     });
     const builtXml = String(builder.build(this.xmlObject));
+    this.xmlObject = undefined;
```
jfeingold35, line 297: Why does this explicitly need to be set to undefined?

jfeingold35 (commented):
Left some comments, also there are merge conflicts to resolve.

src/convert/convertContext/nonDecompositionFinalizer.ts
WillieRuemmele: Done — switched to `Promise.all` upfront then iterate the results. No more eslint-disable needed.

src/convert/convertContext/recompositionFinalizer.ts
WillieRuemmele: Yes — the real fix is `deduplicateChildren` (prevents the unbounded accumulation that caused OOM). The sequential loop here was unnecessary since `getXmlFromCache` already bounds memory via caching. Restored `Promise.all`.

src/convert/streams.ts
WillieRuemmele: Two reasons: (1) guards against Node calling `_read()` more than once — the early-return at the top checks for undefined, and (2) releases the reference so GC can reclaim large JSON trees after they're serialized. Added a comment explaining this.

jfeingold35 (approved):

PR #1817, opened 2026-08-10
fix: resolve decomposed Permission Set parents by suffix

MaxHuet:
## Summary

- Preserve the decomposed Permission Set root metadata file when resolving a component directory.
- Filter root candidates by the metadata parent suffix before accepting a directory-name match.
- Add a regression test covering the Beta2 Permission Set layout.

## Context

Decomposed Beta2 Permission Set files can share the parent full name while using child suffixes. The resolver could therefore select a child fragment as the root metadata XML and later generate an invalid `objectSettings` path.

Related issue: forcedotcom/cli#3621

## Validation

- `corepack yarn build`
- `corepack yarn test:only`
- `corepack yarn test:snapshot`

Known unrelated limitation: `test:registry` still reports metadata types missing from the existing registry fixture.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "node": ">=22.0.0"
   },
   "dependencies": {
-    "@salesforce/core": "^9.0.0",
+    "@salesforce/core": "^9.1.4",
     "@salesforce/kit": "^4.0.0",
     "@salesforce/ts-types": "^3.0.0",
     "@salesforce/types": "^1.6.0",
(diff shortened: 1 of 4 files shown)


MaxHuet:
Revalidation du CLA apres association de l'adresse email du commit au compte GitHub.

WillieRuemmele (approved):
thanks for the contribution, keep an eye on the release notes!

MaxHuet:
Thanks! Will do

PR #1826, opened 2026-08-25
fix: support '#' version separator in AiAgentDefinitionVersion @W-23939860@

WillieRuemmele:
## Summary
- **Fix critical bug**: `key.split('#')` in `metadataApiDeploy.ts` breaks when fullNames contain `#` (AiAgentDefinitionVersion uses `AgentName#N` format). Changed to split only on first `#`.
- **Registry support**: Added missing `suffix` and new `versionSeparator` field for AiAgentDefinitionVersion
- **Pipeline support**: Added `parseAgentVersionFullName` parser, `componentSetBuilder` handling, and `filePathGenerator` bundle mapping

## Work Item
@W-23939860@: Add '#' Version separator support to the CLI

## Proof of Work
- Tests: 41 new tests pass (deployMessages + agentVersionParsing), full suite (30471) passes
- Lint: clean (pre-commit hook)
- Type check: clean (`tsc --noEmit`)
- Build: clean (pre-push hook)

## Test plan
- [ ] Retrieve `AiAgentDefinitionVersion:AgentName#1` — should resolve correctly without key splitting errors
- [ ] Deploy a project containing AiAgentDefinitionVersion bundles — deploy messages should map correctly
- [ ] Retrieve with wildcard `AiAgentDefinitionVersion:AgentName#*` — should retrieve all versions
- [ ] Verify existing BotVersion behavior unchanged (regression check)


src/client/metadataApiDeploy.ts
```diff
@@ -470,7 +470,8 @@ const warnIfUnmatchedServerResult =
   (messageMap: Map<string, DeployMessage[]>): void[] =>
     // keep the parents and children separated for MPD scenarios where we have a parent in one, children in another package
     [...messageMap.keys()].flatMap((key) => {
-      const [type, fullName] = key.split('#');
+      const [type, ...nameParts] = key.split('#');
```
jfeingold35: You could use `const [type, fullName] = key.split('#', 2)` to limit it to 2, and then you wouldn't need to do the `join` in the following line.

src/collections/componentSetBuilder.ts
```diff
@@ -516,11 +518,27 @@ const replacePseudoTypes = async (pseudoTypeInfo: {
...
-      // Normalize entries to Type:* format when no name is provided
-      // This handles both "Type" (no colon) and "Type:" (colon with empty name)
-      const fullName = name.join(':').trim();
-      const normalizedEntry = fullName ? rawEntry : `${typeName}:*`;
-      replacedEntries.push(normalizedEntry);
+      const versionSep = registry.getTypeByName(typeName)?.versionSeparator;
+      if (versionSep) {
+        // Type uses a version separator in fullNames (e.g., '#' for AiAgentDefinitionVersion).
+        // The Metadata API respects specific fullNames, so no client-side filtering is needed.
+        const fullName = name.join(':').trim();
```
jfeingold35: Copy of line 536, so you could move it out of the `if-else`.

jfeingold35 (approved):

PR #1833, opened 2026-09-09
fix: prevent symlink traversal during retrieve (partial-delete and writes) @W-24124120@ @W-24138711@

WillieRuemmele:
## Summary
- **Partial-delete path** (W-24124120): Use `lstatSync` instead of `statSync` when checking if component content paths are directories, preventing symlinked content directories from entering the partial-delete flow. Filter out symlinks within content directories before deletion and add defense-in-depth symlink check in `deleteFilePath`.
- **Write/overwrite path** (W-24138711): Add `findSymlinkOnPath` check in `StandardWriter._write()` before every file write and delete during retrieve, covering all metadata types. If any path segment between the project root and the output file is a symbolic link, the operation is rejected with a clear error.
- **Shared utility**: Extract `findSymlinkOnPath` from `StaticResourceMetadataTransformer` to `fileSystemHandler.ts` so it can be reused across the partial-delete, write, and static resource extraction paths.

## Security Context
During `sf project retrieve start`, two distinct I/O paths followed symbolic links without validation:

1. **Partial-delete** — for `supportsPartialDelete` types (e.g. DigitalExperienceBundle), the logic would follow symlinks when enumerating and deleting local-only content. If a content directory was a symlink to an external path, retrieve could delete files outside the Salesforce project.
2. **File writes** — the `StandardWriter` that writes all retrieved metadata to disk would follow symlinks in the output path. A symlinked file or directory inside the project (e.g. planted in a shared Git repo) could redirect writes to overwrite files outside the workspace.

Both attack vectors require a crafted project (e.g. a malicious Git repo with preserved symlinks) that a victim clones and retrieves against.

@W-24124120@ @W-24138711@

## Proof of Work
- Tests: 1103+ passing, 0 failing
- Lint: clean
- Type check: clean
- New tests: 11 total (4 in `retrieveExtract.test.ts`, 3 in `streams.test.ts`, 4 in `fileSystemHandler.test.ts`)

## Test plan
- [ ] Verify partial-delete still works correctly for non-symlinked DigitalExperienceBundle content
- [ ] Verify symlinked content directories are excluded from partial-delete processing
- [ ] Verify symlinked files within content directories are not deleted
- [ ] Verify normal retrieve writes work for all metadata types (no false positives)
- [ ] Verify retrieve rejects writes through symlinked files in the project
- [ ] Verify retrieve rejects writes through symlinked directories in the project


WillieRuemmele:
## QA Instructions

### Setup
1. Create a Salesforce project with a DigitalExperience site (or use an existing one)
2. Retrieve the site: `sf project retrieve start -m DigitalExperienceBundle`
3. Verify you have a content directory under `force-app/main/default/digitalExperiences/site/<siteName>/`
4. Also have at least one Apex class or other metadata type for the write-path tests

---

### Test 1: Symlinked content directory is NOT followed during partial-delete (W-24124120)
1. Create an external directory with some files:
   ```bash
   mkdir -p /tmp/external-target
   echo "important data" > /tmp/external-target/file1.txt
   echo "critical config" > /tmp/external-target/file2.txt
   ```
2. Replace a content directory with a symlink to the external directory:
   ```bash
   # Pick a content dir, e.g. a view under the site bundle
   CONTENT_DIR="force-app/main/default/digitalExperiences/site/<siteName>/sfdc_cms__view/<viewName>"
   rm -rf "$CONTENT_DIR"
   ln -s /tmp/external-target "$CONTENT_DIR"
   ```
3. Run a retrieve that would trigger partial-delete:
   ```bash
   sf project retrieve start -m DigitalExperienceBundle
   ```
4. **Expected**: The symlinked directory is skipped — files in `/tmp/external-target/` are NOT deleted. The retrieve completes without errors.

### Test 2: Normal partial-delete still works (no false positives)
1. Start with a clean retrieve of a DigitalExperience site (no symlinks)
2. Add a local-only file inside a content directory:
   ```bash
   echo "local only" > "force-app/main/default/digitalExperiences/site/<siteName>/sfdc_cms__view/<viewName>/localFile.txt"
   ```
3. Run: `sf project retrieve start -m DigitalExperienceBundle`
4. **Expected**: The local-only file (`localFile.txt`) IS deleted as part of normal partial-delete behavior — the fix does not break the existing flow for real directories.

---

### Test 3: Symlinked file blocks retrieve writes (W-24138711)
1. Create an external file to serve as the symlink target:
   ```bash
   echo "external secret" > /tmp/external-secret.txt
   ```
2. Replace an Apex class file with a symlink:
   ```bash
   APEX_FILE="force-app/main/default/classes/MyClass.cls"
   rm "$APEX_FILE"
   ln -s /tmp/external-secret.txt "$APEX_FILE"
   ```
3. Retrieve that class:
   ```bash
   sf project retrieve start -m ApexClass:MyClass
   ```
4. **Expected**: The retrieve fails with an error containing "symbolic link" — the external file is NOT overwritten. Verify `/tmp/external-secret.txt` still contains "external secret".

### Test 4: Symlinked directory blocks retrieve writes (W-24138711)
1. Create an external directory:
   ```bash
   mkdir -p /tmp/external-dir
   echo "do not overwrite" > /tmp/external-dir/important.txt
   ```
2. Replace the classes directory with a symlink:
   ```bash
   mv force-app/main/default/classes force-app/main/default/classes.bak
   ln -s /tmp/external-dir force-app/main/default/classes
   ```
3. Retrieve Apex classes:
   ```bash
   sf project retrieve start -m ApexClass
   ```
4. **Expected**: The retrieve fails with an error containing "symbolic link" — nothing is written to `/tmp/external-dir/`. Verify `important.txt` is unchanged.
5. Restore the real directory:
   ```bash
   rm force-app/main/default/classes
   mv force-app/main/default/classes.bak force-app/main/default/classes
   ```

### Test 5: Normal retrieve writes work (no false positives)
1. Ensure no symlinks exist in the project directory
2. Run: `sf project retrieve start -m ApexClass`
3. **Expected**: Retrieve completes normally. All files are written as expected.

src/client/retrieveExtract.ts
```diff
@@ -236,10 +237,22 @@ const isForceIgnored =
...
+    return false;
+  }
+};
+
 const deleteFilePath =
   (logger: Logger) =>
   (fr: FileResponseSuccess): FileResponseSuccess => {
     if (fr.filePath) {
+      if (isSymlinkSync(fr.filePath)) {
+        logger.debug(`Skipping delete of symlink ${fr.filePath} to prevent modification of files outside the project.`);
```
iowillhoit, line 253: `logger.warn` might me more appropriate for this.

iowillhoit:
**🔴 Partial-delete still deletes outside the project when a bundle *ancestor* dir is a symlink (W-24124120 not fully closed)**

The **delete** path only guards the **leaf** (the last segment of the path), so the containment boundary can still be escaped by a symlinked **ancestor** directory.

**The gap**
- `handlePartialDeleteMerges` runs *before* the convert/write step, and its symlink guards are leaf-only:
- `supportsPartialDeleteAndHasContent` → `lstatSync(comp.content)` only refuses to follow the final segment
- the candidate filter and `deleteFilePath` → `isSymlinkSync(join(contentPath, fileName))` only checks the direct child

Neither inspects an **ancestor** of the content root. Local resolution uses `statSync().isDirectory()` (`src/resolve/treeContainers.ts:119`), which *follows* symlinks, so a DEB whose `digitalExperiences` ancestor is a symlink to an external directory resolves normally, its children end in real-file leaves, and partial-delete `rmSync`s the external, local-only files through it.

The write-path guard *would* catch the symlinked ancestor (it walks all segments), but it throws only later in `StandardWriter._write`. By then the delete has already happened and isn't rolled back.

**Repro (confirmed against this branch)**

End-to-end through the real `extract`: a source DEB project where `digitalExperiences/` is relocated outside the project and replaced with a symlink; a retrieve zip missing one content file so it's treated as local-only.

```
victim exists BEFORE : true
extract threw        : Error_retrieve_symlinkError   // write guard fires…
victim exists AFTER  : false                          // …but the external file is already gone
symlink still present: true
```

The delete precedes the write abort, so the write-path guard does not protect against this. (I have a failing mocha I can share)

**Suggested fix**

Give the delete path the same full-path protection the write path already has. In `deleteFilePath` (or as the filter feeding it), reject any candidate where an ancestor segment is a symlink: e.g. `await findSymlinkOnPath(packageRoot, fr.filePath)` is truthy, or do a `realpath`-based containment check against the package root before `rmSync`. That gives delete/write parity and matches the PVR's explicit ask: *"Resolve the real path of every local deletion candidate and verify it remains inside the expected package/project root before any delete operation."*

iowillhoit:
**Containment scope is too narrow, prevents deletes across multiple package dirs**

- **Before:** partial-delete ran `rmSync` on the local component's resolved `contentPath` with no boundary, so it deleted from any of the Project's package dirs (including outside the project 🙅).
- **This change** scopes deletes to `pkg.outputDir`. While the retrieve `output` dir is a single package dir, not the project root (e.g. `sf project retrieve start` uses the default package's path, `start.ts:581`), the local components being retrieved are resolved from the whole project and can live in any package dir. A component in a non-default package now falls outside the boundary and its delete is skipped.

Multiple packageDirectories is a supported, first-class project config (2GP/unlocked packages, modular repos). A DEB/ExperienceBundle in a non-default package dir gives `path.relative(output, filePath)` a `../force-app-2/foo` result. `findSymlinkOnPathSync` treats any leading `..` as out-of-bounds and returns truthy, so `deleteFilePath` skips the `rmSync`. The legit delete is silently skipped, a regression from prior behavior.

**Fix:** anchor containment at the **project root** (all `packageDirectories`) rather than `pkg.outputDir`. That keeps the "nothing deleted outside the project" guarantee while still allowing cross-package deletes.

iowillhoit:
## Follow-up: the two symlink guards use different roots, leaving a write-path escape and a silent delete skip

Verified end-to-end against this branch (local build wired into the installed CLI, real scratch org, source-format merge retrieve of a `LightningComponentBundle`). Both findings share one root cause: the delete-path guard and the write-path guard anchor their containment check at **different roots**.

- Delete path roots at the **project root**: `packageRoot: mainComponents?.projectDirectory ?? pkg.outputDir` (`src/client/retrieveExtract.ts:126`).
- Write path roots at the **package dir**: the writer is `new StandardWriter(output.defaultDirectory)` (`src/convert/metadataConverter.ts:199`), and `defaultDirectory` is `pkg.outputDir`. The guard is `findSymlinkOnPath(this.rootDestination, info.output)` (`src/convert/streams.ts:171`).

Since `projectDirectory` is an ancestor of `pkg.outputDir`, and `findSymlinkOnPath`/`findSymlinkOnPathSync` never lstat the root segment itself (they walk `path.relative(root, destination)` downward), there is a band of segments (the package dir itself and anything between it and the project root) that the delete guard checks but the write guard does not.

### Finding 1 (write-path escape): a symlink at or above `pkg.outputDir` is not caught, and the write escapes the project

If the package directory itself is a symlink pointing outside the project (a plausible shape for a cloned/shared repo), the write-path guard misses it and writes the org's files through the symlink, outside the declared project root, with no error.

Repro: symlink `force-app` to a directory outside the project, then `sf project retrieve start -m "LightningComponentBundle:paginator"`.

Observed:
- `Status: Succeeded` (no `MetadataTransferError`).
- The org's `paginator.js` overwrote the local copy at the external target (a local marker edit was gone afterward), confirming the write landed outside the project.

This is an incomplete fix rather than a regression: before this PR there was no write guard at all. The PR closes the below-`outputDir` case but leaves the at/above-`outputDir` case open.

Direct proof of the asymmetry (same file, two roots):
```
delete-guard root = project root:  SYMLINK CAUGHT
write-guard  root = force-app    :  MISSED -> write proceeds
```

Suggested fix: give the write-path symlink check the same project-root anchor the delete path now uses. Pass `projectDirectory` for the containment check (independent of the `rootDestination` used for path-building and forceignore), or resolve realpath and verify containment against the project root. Both guards should agree on one boundary.

### Finding 2 (observability): the delete-path skip has no reachable log

When the delete guard suppresses a partial-delete, there is no signal, not even at `--dev-debug`. The only log lives in `deleteFilePath`:
```ts
const symlink = findSymlinkOnPathSync(packageRoot, fr.filePath);
if (symlink) {
  logger.debug(`Skipping delete of ${fr.filePath} — path segment ${symlink} is a symbolic link.`);
  return fr;
}
```
But a symlinked candidate is removed by a pre-filter in `handlePartialDeleteMerges` before it can reach `deleteFilePath`:
```ts
.filter((fileName) => !findSymlinkOnPathSync(packageRoot, path.join(matchingLocalComp.contentPath, fileName)))
```
So the `debug` line is effectively dead for the real symlink case (it would only fire on a TOCTOU race where the path became a symlink between the filter and the delete). A 1,200-line `--dev-debug` capture of the repro contained no skip/symlink line at all.

Combined effect in the symlinked-package-dir case: a "Succeeded" retrieve that quietly wrote outside the project and quietly declined a delete, with nothing in the logs pointing at the symlink.

Suggested fix: log at the point of suppression (in the pre-filter) so a skipped partial-delete is visible, and reconcile the loud-throw (write) vs silent-skip (delete) behaviors so the two paths signal consistently.

project-root-can-be-symlink-and-write-still-follows

iowillhoit (approved):

